PR #5, opened 2023-05-23
A GCH example

ceriottm:
This is pretty much the same example as in skmatter, but here features are computed dynamically using rascaline, and the chemiscope export is commented on a bit more in depth. At present this does not work because skmatter on pip is outdated.


agoscinski:
chemiscope does not work with the sphinx examples for the moment https://github.com/lab-cosmo/software-cookbook/issues/2 I wrapped around the `chemiscope.show` a if case checking if it is run in an environment that can display chemiscope.

PR #10, opened 2023-05-26
Added PyTorch Geometric introduction tutorial 

abmazitov:


examples/torch/introduction_to_torch_geometric_for_atomistic_machine_learning.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+######################################################################
+# This notebook explains the basics of using Pytorch Geometric library for
+# atomistic machine learning. It is assumed that you are familiar with the
+# basics of PyTorch. If not, please check out the `official PyTorch
+# tutorials <https://pytorch.org/tutorials/>`__.
+# 
+# For the following work we need to first install the PyTorch and PyTorch
+# Geometric libraries. YThere are several options how to do it, the
```
agoscinski:
```suggestion
# Geometric libraries. There are several options how to do it, the
```

examples/torch/introduction_to_torch_geometric_for_atomistic_machine_learning.py
```diff
@@ -0,0 +1,230 @@
...
+
+######################################################################
+# The way it works is that it joins all graphs in the mini-batch into a
+# single large graph. The connectivity information for this new graph is
+# stored in a new block diagonal adjacency matrix. The node and edge
+# features are stored in a single large node feature tensor and edge
+# feature tensor, respectively.
+# 
+# I.e., if we had a graph :math:`G_1 = (X_1, A_1)` and graph
+# :math:`G_2 = (X_2, A_2)`, then uniting them into a Batch will result in
```
agoscinski, line 194: I feel like this part is unnecessary abstract explanation for "We stack graph data together and store an index for each graph/structure".

agoscinski (commented):
I think this tutorial can be more concrete about atomic structures. So including a neighborlist that has distances (edge information) and positions (node information) using ase would be nice. 

I thinks it is important to note in the text where exactly the torch_geometric data utility helps. We only need to define how to build a graph out of a structure (e.g. neighborlist or half neighborlist) and torch_geometric takes over the job how to make a batch or mini-batches out of this. 

Below are more suggestions, that I extend this tutorial to, but I would put them into an issue for further improvements of this tutorial. If you could address the above comments, then I would merge the current version. If you do edits on the .ipynb or .py file, then I can take over the rest.

### torch_geometric transform
One could also show how different graphs can be in principle created with different information (species or distances or distance vectors) by using different transformers as in torch_alchemical and by that we can easily filter out nodes/edges (e.g. atoms that correspond to a certain species)

### torch_geometric Datasets

#### pretransform
Also using the pretransfrom functionality of custom datasets might be an useful for the usecase that I you don't want to dynamically compute the neighborlist for training over epochs so you precompute it You might want to do the same for features if you don't care about forces, or if you have a autograd interface like we have with rascaline.

#### InMemoryDataset vs Dataset
 Having everything in memory or loading it dynamically, both have valid use cases.

PR #11, opened 2023-05-26
Activating github actions for all PRs

agoscinski:


agoscinski:
Just realized that the Github action is not needed. Will delete in this PR #12

PR #27, opened 2023-10-12
Few tweaks to roy example and re-trigger RTD build

ceriottm:
As the title says. RTD build and deployment failed, and we want to debug. Also improving a bit the chemiscope output for the ROY example

----
:books: Documentation preview :books:: https://software-cookbook--27.org.readthedocs.build/en/27/


Luthaf:
This should now run fine, if you can rebase on top of master!

PR #29, opened 2023-10-17
Use separate conda env for each example

Luthaf:
To merge after #28.

This allows us to have separate versions of dependencies for different examples, and pull in non-python dependencies such as dftb.

This also allows to use different versions of Python itself (librascal does not work with Python > 3.10).

----
:books: Documentation preview :books:: https://software-cookbook--29.org.readthedocs.build/en/29/


docs/src/lode_linear/images/sphx_glr_lode_tutorial_001.png
PicoCentauri, line 1: I think they should not be in the repo.

examples/lode_linear/environement.yml
```diff
@@ -0,0 +1,15 @@
+name: lode_linear
+dependencies:
+  - python=3.11
+  - pip
+  - pip:
+    # sphinx-gallery and dependencies
+    - sphinx
+    - pillow
+    - sphinx-gallery
```
PicoCentauri: Since these are used for every example, can't we put them in the tox file?

examples/lode_linear/environement.yml
```diff
@@ -0,0 +1,15 @@
...
+    - matplotlib
```
PicoCentauri: not really a sphinx-gallery example. We could have example that do not have to plot anything.

PicoCentauri (changes requested):
Very nice improvement. I think the could try to cleanup the `enviroment.yml` files a bit more. i.e. move general deps to the tox  file if possible.

examples/lode_linear/environement.yml
Luthaf: We could install them from inside `tox.ini` yes

examples/lode_linear/environement.yml
Luthaf: everything is a sphinx-gallery example, since that's what we use to transform the python script to rst

PicoCentauri (approved):

PR #34, opened 2024-02-15
Add artifacts upload for website

PicoCentauri:
Since most likely read the docs is not working for this repository for now we store the build website data as an artifact. which can be downloaded.

I will add a documentation for this within this PR once we are sure that this workflow is working.


PicoCentauri:
Okay it seems to work. We can merge this and check if the link is posted correctly. 🤞

PR #35, opened 2024-02-16
Cleaned up and converted to rascaline the selection example

ceriottm:
Now that CI and local kind of work, we can start making the examples nicer and advertise them a bit.

### DIFF
--- a/examples/sample_selection/README.rst
+++ b/examples/sample_selection/README.rst
@@ -1,7 +1,7 @@
 Sample and Feature Selection
 ============================
 
-This is an example of the selection of a diverse subset of structures 
+This is an example of the selection of a diverse subset of structures
 from a dataset of candidates (a collection of barium titanate structures)
 using SOAP features and FPS and CUR selection. It then also demonstrates
 how the same techniques can be used to select the most relevant features.
--- a/examples/sample_selection/README.rst
+++ b/examples/sample_selection/README.rst
@@ -1,3 +1,12 @@
-Sample Selection
-================
+Sample and Feature Selection
+============================
+
+This is an example of the selection of a diverse subset of structures 
+from a dataset of candidates (a collection of barium titanate structures)
+using SOAP features and FPS and CUR selection. It then also demonstrates
+how the same techniques can be used to select the most relevant features.
+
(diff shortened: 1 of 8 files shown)


examples/sample_selection/README.rst
```diff
@@ -1,3 +1,12 @@
...
+============================
+
+This is an example of the selection of a diverse subset of structures
+from a dataset of candidates (a collection of barium titanate structures)
+using SOAP features and FPS and CUR selection. It then also demonstrates
+how the same techniques can be used to select the most relevant features.
+
+It showcases the use of ``rascaline`` to compute descriptors for a
+database of atomic structures, and ``scikit-matter`` to compute the
+FPS and CUR selections. The structures are visualized using ``chemiscope`` widgets.
```
PicoCentauri, line 11: I think unfortunately this nowhere rendered for now.

Luthaf: We should put this in the header of the example maybe?

PicoCentauri: But then this is a bit duplicated, no? The problem is that we somehow abusing the structure of the gallery. I think the idea is that several example scripts are within one folder while we always only have one.

Luthaf: for sure we are abusing the structure a bit here =). What I meant was to have an empty README, and push the content to the example if it is important

DavideTisi: The README is not rendered for any example I thought it was more of a reminder for someone that want to modify the directory to understand what it is.

I do not know for me it is fine both having it empty and putting this in the `*py` file or keep it here and not have it rendered.

A part for this for me the PR is fine

ceriottm: I was considering this more as a reminder for developers - it's true that the notebook format is self-documenting but I like for the README to say something

DavideTisi: I agree with michele if is ok for the others I would merge

DavideTisi (approved):

examples/sample_selection/README.rst
Luthaf: Works for me!

PR #37, opened 2024-02-16
Start workflows every now and then

DavideTisi:
The workflow runs every Monday at 8am UTC.
I hope it is correct I followed [this guide](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#scheduled-events-schedule) 
So far, only `lint`and `docs` run regularly. Do we want also the artefacts upload to do the same?


PicoCentauri:
Yes, looks wonderful!

PR #39, opened 2024-03-25
Run each example in one CI job

PicoCentauri:
This should speed up our CI. Jobs for each examples store their results in a Github artifact. The actual docs job is just collecting these artifacts and generate the final documentation page.

I also changed some minor other infrastructure things. Most notable:

 - change name of `environement.yml` to the more common `enivornment.yml`
 - set `matplotlib` dependency for each example since this might not be used by all examples
 - nake the datasets of the LODE and the sample selection available for download


PicoCentauri:
The CI so fast now! 🚀

.github/workflows/docs.yml
```diff
@@ -10,27 +10,71 @@ on:
     - cron: '0 8 * * 1' # run every Monday at 8am UTC
 
 jobs:
+  generate-example:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example-name: [lode_linear, roy_gch, sample_selection, gaas_map, cp2k_run_batch]
```
Luthaf:
```suggestion
        example-name: 
          - lode_linear
          - roy_gch
          - sample_selection
          - gaas_map
          - cp2k_run_batch
```

PicoCentauri: ahh yes sure! Thanks!

PR #41, opened 2024-03-25
Use nox instead of tox

Luthaf:
This gives us a lot more flexibility in the workflow. 

Now, `nox -e lode-linear` will execute the example and then build the website locally, including only this example; and `nox -e docs` will build everything.

`nox -e lint` and `nox -e format` work the same.

Conda dependencies should now be available inside the virtualenvs, meaning we don't need to install rust or cp2k in CI ourself.

---

I had to deactivate the cp2k example, since it was not running on my computer. I got some error about pseudopotential files not being found. I think we can merge this once CI is happy and try to fix cp2k tomorrow.


examples/roy-gch/environment.yml
```diff
@@ -1,11 +1,13 @@
...
 dependencies:
   - python=3.11
   - pip
+  - rust
   - pip:
     - ase
     - chemiscope
     - matplotlib
-    - skmatter
     - metatensor
```
PicoCentauri, line 11: I was using alphabetical order here, as it is done by the our linters in the Python files. Then at least we have some consistency.

noxfile.py
```diff
@@ -0,0 +1,182 @@
...
+        output.write(".. toctree::\n")
+        output.write("    :caption: Table of Contents\n")
+        output.write("    :maxdepth: 1\n")
+        output.write("\n")
+
+        for file in glob.glob("docs/src/examples/*/*.rst"):
+            if os.path.basename(file) != "sg_execution_times.rst":
+                path = file[9:-4]
+
+                output.write(f"    {path}\n")
```
PicoCentauri: Good idea adding this automatically.

noxfile.py
```diff
@@ -0,0 +1,182 @@
...
+        session.install("-r", requirements)
+
+    with open("docs/src/index.rst", "w") as output:
+        with open("docs/src/index.rst.in") as fd:
+            output.write(fd.read())
+
+        output.write(".. toctree::\n")
+        output.write("    :caption: Table of Contents\n")
+        output.write("    :maxdepth: 1\n")
+        output.write("\n")
```
PicoCentauri: This seems to be static. Should we write this directly in `index.rst.in`?

PicoCentauri (commented):
Cool! Once our cp2k mess is working as well this seems like a very useful improvment, which once ensure that this works we should port to other COSMO projects as well.

noxfile.py
Luthaf: 👍

PR #44, opened 2024-03-26
Re-enable cp2k example

Luthaf:


Luthaf:
So the only solution to install cp2k everywhere was to install it manually with the right tool (`conda`/`brew`) inside the script.

examples/batch-cp2k/reference-trajectory.py
```diff
@@ -136,17 +148,25 @@ def write_cp2k_in(
...
+# installation, this might not be necessary!
 
 
-def mkdir_force(*args, **kwargs) -> None:
-    """Warpper to ``os.mkdir``.
+def download_parameter(file):
+    path = os.path.join("parameters", file)
+
+    if not os.path.exists(path):
+        url = f"https://raw.githubusercontent.com/cp2k/cp2k/master/data/{file}"
```
PicoCentauri: Should we maybe not use the master but rather a fixed tag? I don't expect that they change how basis sets are stored but just to be sure.

```suggestion
        url = f"https://raw.githubusercontent.com/cp2k/cp2k/support/v2024.1/data/{file}"
```

PicoCentauri (commented):
Very nice! One quick comment.

PR #46, opened 2024-03-26
Infrastructure updates

### LINKED ISSUE #42: Let users download the `environment.yml` files that correspond to the given examples

### LINKED ISSUE #43: List authors and (contactable) maintainers for each cookbook example

Luthaf:
Fix #42 
Fix #43

### DIFF
--- a/examples/batch-cp2k/reference-trajectory.py
+++ b/examples/batch-cp2k/reference-trajectory.py
@@ -2,7 +2,8 @@ r"""
 Batch run of CP2K calculations
 ==============================
 
-.. start-body
+:Authors: Matthias Kellner `@bananenpampe <https://github.com/bananenpampe/>`_,
+          Philip Loche `@PicoCentauri <https://github.com/PicoCentauri/>`_
 
 This is an example how to perform single point calculations based on list of structures
 using `CP2K <https://www.cp2k.org>`_ using its `reftraj functionality
--- a/noxfile.py
+++ b/noxfile.py
@@ -128,6 +128,30 @@ def build_docs(session):
 
                 output.write(f"   {path}\n")
 
+                # TODO: Explain
+                with open(file) as fd:
+                    content = fd.read()
+
+                if "Download Conda environment file" in content:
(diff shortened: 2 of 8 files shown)


examples/batch-cp2k/reference-trajectory.py
```diff
@@ -2,7 +2,8 @@
 Batch run of CP2K calculations
 ==============================
 
-.. start-body
+:Authors: Matthias Kellner `@bananenpampe <https://github.com/bananenpampe/>`_,
+          Philip Loche `@PicoCentauri <https://github.com/PicoCentauri/>`_
```
jwa7, line 6: Should we separate authors and maintainers, a la @PicoCentauri suggestion in #43 ? In a similar vein to metatensor-models, it might be better to just have one maintainer that is responsible

PicoCentauri: Talkes to @Luthaf and we agreed that authors should be enough.

Tutorials should not be super hatd to maintain. With our conda env they should basically not break 😅

jwa7: Ok fair enough!

jwa7 (approved):

noxfile.py
```diff
@@ -128,6 +128,30 @@ def build_docs(session):
...
+                                fd.write(
+                                    """
+    .. container:: sphx-glr-download
+
+      :download:`Download Conda environment file: environment.yml <environment.yml>`
+"""
+                                )
+
+                            fd.write(line)
+                            fd.write("\n")
```
PicoCentauri, line 153: Seems to work very well. I don't know if like that this is here. I rather would like to put this as an extensions sphinx-gallery directly.

I will explore this.

Luthaf: Did you find a way to do this? Or should we merge this version for now?

PicoCentauri: I have an idea how to do this within the conf.py. But, we can merge this and I change this later!

PicoCentauri (approved):

PR #47, opened 2024-03-26
Improve Feature Selection

HannaTuerk:
Hello,
I updated the sample selection example to use equisolve (before we used skmatter). I also kept an example for feature selection directly with skmatter at the bottom.


DavideTisi (commented):
For me it is fine, I had to build it on my laptop. it cannot build the cp2k example but the sample selection seems ok

DavideTisi (approved):

PR #60, opened 2024-05-05
added nox -e clean

DavideTisi:
added nox -e clean to clean all the temporary files in examples.

followed @ceriottm discussion this is a proposal for implementing the clean function in nox.
it uses `git ls-files --other example` to check all the files not git tracked in the example dir


noxfile.py
```diff
@@ -249,3 +263,16 @@ def format(session):
...
+
+
+@nox.session
+def clean(session):
+    """Remove temporary files"""
+    tracked_files_output = get_example_other_files()
+    for ifile in tracked_files_output:
+        os.remove(ifile)
+    flist = glob.glob("examples/*")
+    # check for empty list and remove them
```
PicoCentauri:
```suggestion
    # Remove empty folders
```

noxfile.py
```diff
@@ -249,3 +263,16 @@ def format(session):
     session.run("black", *LINT_FILES)
     session.run("blackdoc", *LINT_FILES)
     session.run("isort", *LINT_FILES)
+
+
+@nox.session
+def clean(session):
+    """Remove temporary files"""
+    tracked_files_output = get_example_other_files()
+    for ifile in tracked_files_output:
```
PicoCentauri:
```suggestion
    for ifile in get_example_other_files():
```

noxfile.py
```diff
@@ -249,3 +263,16 @@ def format(session):
...
+        os.remove(ifile)
+    flist = glob.glob("examples/*")
```
PicoCentauri, line 326: Is glob only listing folders or also files here?

noxfile.py
```diff
@@ -249,3 +263,16 @@ def format(session):
     session.run("black", *LINT_FILES)
     session.run("blackdoc", *LINT_FILES)
     session.run("isort", *LINT_FILES)
+
+
+@nox.session
+def clean(session):
```
PicoCentauri: I don't know if you want but you can may also want to delete the folder "rendered folder" "docs/src/examples".

noxfile.py
DavideTisi: Also files but then you check the len, it should skip files since they have len>0

noxfile.py
DavideTisi: I did not thought about that you can add that, sure

noxfile.py
```diff
@@ -73,6 +73,20 @@ def get_example_files():
     return [folder + "/" + file for file in filtered_files]
 
 
+# We want to mimic
+# git ls-files --other examples
+def get_example_other_files():
```
Luthaf: This feels extremely similar to `git clean -xd -- <folder>` (check `git clean -h`)

DavideTisi: I have no idea what that command is, so probably yes

noxfile.py
ceriottm: I actually was thinking of removing _specifically_ that folder, and not stray files in `examples/` that could be for instance test jupyter notebooks or similar. it is the stray files in docs/src/examples/ that generate the most problems in my experience

ceriottm:
With @Luthaf  we were discussing of how we should call this target. clean-build seemed a good descriptive one. I'd also add docs/build, and .nox if you haven't added those already.

DavideTisi:
I thought we did not want to add `.nox`

ceriottm:
OK, I went through this and as far as I'm concerned, I'm happy. I separated clean_build from clean_examples, as I think people may want to keep experiments in examples sometimes, but need to clean the build. Maybe this approach gives us more flexibility to keep stuff we don't want to remove for some reason.

DavideTisi:
for me it is fine

DavideTisi:
@ceriottm do we want to merge this?
@Luthaf @PicoCentauri what do you think ?

ceriottm:
Lgtm, I think I cleaned it up a while ago

noxfile.py
```diff
@@ -74,6 +74,20 @@ def get_example_files():
     return [os.path.join(folder, file) for file in filtered_files]
 
 
+# We want to mimic
+# git ls-files --other examples
+def get_example_other_files(fd):
+    folder = os.path.join(os.getcwd(), fd)
+    # Get the list of ignored files
+    # Get the list of all not tracked files
```
Luthaf: comment seems duplicated

noxfile.py
```diff
@@ -74,6 +74,20 @@ def get_example_files():
     return [os.path.join(folder, file) for file in filtered_files]
 
 
+# We want to mimic
+# git ls-files --other examples
```
Luthaf:
```suggestion
```

noxfile.py
```diff
@@ -293,3 +307,27 @@ def format(session):
...
     for file in LINT_FILES:
         remove_trailing_whitespace(file)
+
+
+@nox.session
+def clean_build(session):
+    """Remove temporary files and building folders"""
+
+    # remove building folders and the nox cache
+    for i in ["docs/src/examples/", "docs/build", ".nox"]:
```
Luthaf: why `.nox`? this will remove itself when running =/

noxfile.py
```diff
@@ -293,3 +307,27 @@ def format(session):
...
+
+    # remove all untracked files from examples
+    for ifile in get_example_other_files("examples"):
+        os.remove(ifile)
+
+    flist = glob.glob("examples/*")
+    # Remove empty folders
+    for fl in flist:
+        if 0 == len(glob.glob(fl + "/*")):
+            os.rmdir(fl)
```
Luthaf:
```suggestion
    for path in flist:
        if len(glob.glob(os.path.join(path, "*"))) == 0:
            os.rmdir(path)
```

noxfile.py
DavideTisi: @ceriottm I think you added that piece

DavideTisi: Even without michele comment I kind of agree that removing folders like `.nox` is the point of this function

Luthaf: hmm, I see this more as a way to clean temporary files created by the example, not to clean the virtualenvs. Also `rm -rf .nox` is trivial compared to what `nox -e clean` does, so leaving them separate is not too much of a burden on the end user.

ceriottm: So, the rationale for this whole PR is to have a setup similar to the clean_python script of other projects, wiping all build and caches. Upon checking, it's true that all of those don't wipe .tox so even though I added it in the first place, I now agree that for consistency we should not (or we should do this consistently everywhere). Removing it for now (and merging)

ceriottm (approved):

PR #66, opened 2024-08-10
Sectioned content

ceriottm:
Make content split into sections, and de-branded


ceriottm:
also considering renaming the repo to `atomistic-cookbook` which is less generic

PicoCentauri (commented):
I am fine with renaming even though the name "atomistic simulations" is not covering everything that we do in here, right?

PR #70, opened 2024-08-22
Path integral metadynamics

ceriottm:
A tutorial to run PI + METAD simulations with i-PI + PLUMED


ceriottm:
Ready to be checked!

DavideTisi (approved):

PR #74, opened 2024-09-08
Miscellaneous clean-ups of several examples

ceriottm:
Adding backreferences to papers, reordering the build scripts, generally trying to improve the repo


src/generate-gallery.py
```diff
@@ -7,7 +7,7 @@
 from chemiscope.sphinx import ChemiscopeScraper
 
 
-HERE = os.path.realpath(os.path.dirname(__file__))
+HERE = os.path.realpath(os.path.join(os.path.dirname(__file__), "../"))
```
Luthaf:
```suggestion
ROOT = os.path.realpath(os.path.join(os.path.dirname(__file__), "../"))
```

Would be a better name

ceriottm: hey I was just using the name that had been used ^^

Luthaf (approved):

PR #78, opened 2024-09-18
Use `plt.show` to avoid extra output

frostedoyster:
As in the title


ceriottm (approved):
We should probably provide some guidelines to have cleaner examples. Thanks for fixing these.

PR #80, opened 2024-09-26
Stub of an o3-averaging example

ceriottm:
Examples of how to highlight and correct the impact of non-equivariant energy and force predictions, e.g. when using an unconstrained ML potential.


ceriottm:
THIS IS A DRAFT PR FOR THE MOMENT

PR #82, opened 2024-09-27
Restructure the cookbook with thematic and software sections

ceriottm:
Better structure allowing for overlap between topics and classification based on used tools.

PR #84, opened 2024-10-04
Add generated files to .gitignore

Luthaf:
This is a follow up to #82, to prevent the automatically generated .rst files from being committed.


ceriottm (approved):
LGTM. There's a lot of text reflow noise but nothing that hurts/

PR #87, opened 2024-10-07
Restructure the docs folder in a more standard way

ceriottm:
Remove `latest/` prefix which doesn't really make sense here and improve the sitemap by adding also the root page explicitly. Also fix the canonical for the main page so it's clearer for indexing.


.github/workflows/docs.yml
```diff
@@ -90,9 +90,9 @@ jobs:
           if [[ "$REF_KIND" == "tags" ]]; then
               TAG=${GITHUB_REF#refs/tags/}
               mv ../docs/build/html $TAG
-          else
-              rm -rf latest
-              mv ../docs/build/html latest
+          else # deploys directly to the root of the website
```
Luthaf: I would remove the code handling git tags here, since the "normal" path would remove the tag on the next build anyway.

I can't do a suggestion, but this would just be replacing the `if` block with only

```
rm -rf *
mv ../docs/build/html/* .
```

ceriottm: done, and also moved the post-build stuff to conf.py.
will merge now thanks for the feedback

PR #89, opened 2024-10-13
Thermostatting tutorial

ceriottm:
An opinionated introductory tutorial for thermostatting


ceriottm:
Still work in progress, but I want to have CI tests run

ceriottm:
Added a few of you as reviewers, I'd like to merge this rather quickly so if one of you has a bit of bandwitdh we can merge and then let others comment, if and when they have time, in a separate PR

examples/thermostats/README.rst
```diff
@@ -0,0 +1,9 @@
+Thermostattin molecular dynamics simulations, a primer
```
DavideTisi:
```suggestion
Thermostatting molecular dynamics simulations, a primer
```

examples/thermostats/thermostats.py
```diff
@@ -0,0 +1,900 @@
...
+# rapidly to the target value. However, it takes a bit longer
+# (approximately 0.5 ps) to reach equipartition between O and H
+# atoms. This is an important shortcoming of global thermostats:
+# since they only target the total kinetic energy, they must rely
+# on internal energy redistribution to reach equilibrium between
+# different degrees of freedom.
+# Liquid water is a very ergodic system, in which all degrees of
+# freedom are strongly coupled, so this is not a major issue. However
+# care must be taken when modeling a quasi-harmonic crystal, or a
+# molecular system in which the coupling between molecules is weaker.
```
DavideTisi: @ceriottm do you have a concrete example of this, a system for which this thermostat does not work well.

examples/thermostats/thermostats.py
```diff
@@ -0,0 +1,900 @@
...
+ax.legend()
+plt.show()
+
+
+# %%
+# The velocity-velocity autocorrelation function is
+# essentially indistinguishable from the reference, computed
+# with an ensemble of NVE trajectories starting from canonical
+# samples. In fact, the small discrepancies are mostly due to
+# incomplete convergence of the averages in thi short trajectory.
```
DavideTisi:
```suggestion
# incomplete convergence of the averages in the short trajectory.
```

DavideTisi (commented):
some typos

examples/thermostats/thermostats.py
```diff
@@ -0,0 +1,900 @@
...
+# averaging over multiple constant-energy trajectories.
+# As we shall see, this protocol can often be simplified greatly, by choosing
+# thermostats that don't interfere with the natural microscopic dynamics.
+
+# %%
+# Running simulations
+# ~~~~~~~~~~~~~~~~~~~
+#
+# We use `i-PI <http://ipi-code.org>`_ together with a ``LAMMPS`` driver to run
+# all the simulations in this recipe. The two codees need to be ran separately,
```
DavideTisi:
```suggestion
# all the simulations in this recipe. The two codes need to be run separately,
```

DavideTisi:
for me we can merge. The only question (a part from the typos) is that you could put an explicit example where the global thermostat fails (or is slowto converge), even without running the simulation just saying the name of the system if you know one

examples/thermostats/thermostats.py
ceriottm: I thought that the slow equipartition between O and H was a good enough example. 
Adding a harmonic crystal would lengthen a lot an already quite long tutorial. Do you think what we have is not enough? maybe I can distort the initial configuration to make this effect more prominent?

DavideTisi:
I think that just citing some system like you did now+ the example with water is enough. I was just thinking that only the water example was not very strong because the convergence is still quite fast. 
Now for me it is mergable if the CI is happy

DavideTisi (approved):

PR #96, opened 2024-11-12
Rpc mts

DavideTisi:


ceriottm:
Starting to look at this. Can you put in place the same caching mechanism that is used in other i-PI based notebooks so that if you re-run it it does not re-run i-PI? it'll make much faster to tweak the parameters to have a reasonably fast but insightful setup

ceriottm (approved):
Amazing Davide, thanks a lot for converting that old tutorial to this. Very very nice.
I re-read it and tweaked a couple of points, and I'll merge as soon as tests pass.

PR #103, opened 2024-12-10
Free energy

DavideTisi:
Tutorial 3 from https://github.com/i-pi/piqm2023-tutorial/blob/main/03-free_energy_methods/tutorial.ipynb


DavideTisi:
@ceriottm this notebook comes from the IPI school. It has a problem: it's too long, it takes about 18 minutes. Can I simply reduce the timesteps for everything and have results that are not at convergence?

PR #104, opened 2024-12-11
Add example for learning capabilities with torchpme

E-Rum:


DavideTisi:
is it ready for review?

E-Rum:
Yes!

DavideTisi (approved):
I added few modifications to the titles. Look if you like it, if so you can merge. Given that the CI passes.

DavideTisi:
Ah in principle this notebook would be too long. We said no more than 10 min on CI, but the results in the figures really show that the number of epochs is necessary to see the correct behaviour of the charge so for me it is ok

ceriottm:
> Ah in principle this notebook would be too long. We said no more than 10 min on CI, but the results in the figures really show that the number of epochs is necessary to see the correct behaviour of the charge so for me it is ok

Ackshually I think for the paper we showed that picking a better learning rate cut the epochs by 10, @E-Rum can you update so it doesn't take forever?

DavideTisi:
Screenshot 2025-02-10 alle 11 25 07


Just make the X axis the same like put `sharexaxis=True` instead of False

E-Rum:
Ok, now it's only ~3m. Please give a thumbs up to this message if it works now, and I'll merge

PR #105, opened 2024-12-18
Sample selection update to featomic

HannaTuerk:
Updated the FPS example to featomic. I tested the output, it is the same as with rascaline

the commit log shows changes to the hamiltonian-qm7 example, I deleted the folder here before send the PR, so only relevant files get merged


Luthaf (approved):
Looks good to me!

DavideTisi:
Is it normal that to me it fails the tests?

HannaTuerk:
The periodic-hamiltonian example builds fine on my laptop, idk why it fails here. I did not change anything in it. I checked with Paolo, and none of the imported packages were updated, so I am also not sure why this happened

PR #106, opened 2024-12-19
Revert "Sample selection update to featomic"

HannaTuerk:
Reverts lab-cosmo/atomistic-cookbook#105


ceriottm:
Why are you reverting? I thought the problem was unrelated?

HannaTuerk:
Yes, just wanted to make sure it's really unrelated. I'll look into why the periodic ham test fails now

PR #107, opened 2024-12-19
Add PYSCF-AD backend to environment.yml

HannaTuerk:
This fixes the new error the periodic hamiltonian example throws related to jax


HannaTuerk:
ok, nevermind.. this worked on my laptop.. Will investigate this further I guess

HannaTuerk:
ok, I did the quickfix now with #noqa at the relevant imports of mlelec. Not so pretty, but the sphinx-gallery kept ignoring the environment variable after creating the new env

ppegolo (approved):
Good for me, thanks for fixing this!

PR #108, opened 2025-01-19
fix name tag

DavideTisi:
should fix the CI with the nametag change of last ipi PR


DavideTisi:
I also had to pin the scipy version to an older version in roy-gch to make skmatter working.

@ceriottm if this is ok for you, accept the pr that we can merge and fix the ci

ceriottm (approved):

PR #109, opened 2025-01-20
Add example for constructing flexible water models

PicoCentauri:
This PR adds an example for how to build flexible three/four point water models from scratch and how to export and run them with metatensor atomistic models. The electrostatic is handles using torchpme.

The model it self is quite long and therefore located in a different file. This also makes the usage and download easier in my opinion. All important aspects of construct these water models are discussed in the example

# TODO
- [x] make the model run (currently there are still some shape errors)
- [x] add example parameters for spc/e, tip3p and tip4p...


examples/water-model/water-model.py
```diff
@@ -0,0 +1,649 @@
...
+# these in simulations. For periodic systems the Coulomb energy is given by:
+#
+# .. math::
+#
+#  V_\mathrm{Coulomb} = \frac{1}{2} \sum_{i,j} \sideset{}{'}\sum_{\boldsymbol n \in
+#   \mathcal{Z}} \frac{q_i q_j}{\boldsymbol r_{ij} + \boldsymbol n L}
+#
+# The sum over :math:`\boldsymbol n` takes into account the periodic images of the
+# charges and the prime indicates that in the case :math:`i=j` the term :math:`n=0` must
+# be omitted. Further :math:`boldsymbol r_{ij} = \boldsymbol r_i - \boldsymbol r_j` and
```
DavideTisi:
```suggestion
# be omitted. Further :math:`\boldsymbol r_{ij} = \boldsymbol r_i - \boldsymbol r_j` and
```

DavideTisi (approved):

examples/water-model/model.py.example
```diff
@@ -276,14 +276,14 @@ class WaterModel(torch.nn.Module):
         # select pairs within the same molecule
         mol_mask = (i // 3) == (j // 3)
 
-        bond_mask = mol_mask & (species[i] == 8)
+        bond_mask = mol_mask & (species[i] == 8) & (species[j] == 1)
```
PicoCentauri: I think this is not necessary because the mol mask already contains only pairs within a molecule and we only have a single oxygen (there are no O-O) pairs. But we can leave it if you like for being explicit.

examples/water-model/water-model.py
```diff
@@ -47,167 +48,90 @@
...
-        trajectory=True, structure_settings={"unitCell": True}
-    ),
-)
-
-# %%
-#
-# The Model
-# ---------
+# The q-TIP4P/F Model
+# -------------------
```
PicoCentauri, line 64: I somehow liked the idea going from a simple 3 point model to a more complex one. But its fine if this example should be only about 4-point models.

examples/water-model/water-model.py
```diff
@@ -216,16 +140,25 @@
 # where :math:`k_\mathrm{angle}` is the force constant and :math:`\theta_0` is the
 # equilibrium angle between the three atoms. We use the following parameters:
 
-HOH_angle_coefficient = 75.90  # kcal/mol/rad^2
-HOH_equilibrium_angle = 112.0  # degrees
+
+def bend_energy(
```
PicoCentauri, line 161: I like the common word "angles"

examples/water-model/water-model.py
```diff
@@ -236,38 +169,132 @@
...
+#
+# .. math::
+#
+#  V_\mathrm{LJ}(r) = 4 \epsilon \left[ \left( \frac{\sigma}{r} \right)^{12} - \left(
+#  \frac{\sigma}{r} \right)^6 \right]
+#
+# where :math:`\epsilon` is the depth of the potential well and :math:`\sigma` is the
+# distance at which the potential is zero. For water there is only an oxygen-oxygen
+# Lennard-Jones potential. In a typical 4-points water model are no LJ interactions
+# between hydrogen atoms and between hydrogen and oxygen atoms.
```
PicoCentauri: Your formulation is a bit complicated.

Maybe just say there is only an interaction between oxygen and AFAIK this holds not only for 4 point models but also 3 point.

examples/water-model/water-model.py
```diff
@@ -236,38 +169,132 @@
...
+plt.show()
+
+
 # %%
 #
 # Electrostatic Potential
 # ^^^^^^^^^^^^^^^^^^^^^^^
 #
-# Since electrostatic interactions are long-ranged it is not fullt trivial to compute
+# Since electrostatic interactions are long-ranged it is not fully trivial to compute
```
PicoCentauri: Remove the "Since"

examples/water-model/water-model.py
```diff
@@ -236,38 +169,132 @@
...
 # these in simulations. For periodic systems the Coulomb energy is given by:
 #
 # .. math::
 #
 #  V_\mathrm{Coulomb} = \frac{1}{2} \sum_{i,j} \sideset{}{'}\sum_{\boldsymbol n \in
```
PicoCentauri, line 285: I think we should SI units here add the $1/4π\varepsilon_0$ and comment below that `p3m_prefactor` is exactly this factor for kcalmol and Ånsgtrom

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+    o_pos = positions[::3]
+    h1_pos = positions[1::3]
+    h2_pos = positions[2::3]
+
+    oh_dist = torch.concatenate([
+            torch.sqrt(((h1_pos-o_pos)**2).sum(axis=1)),
+            torch.sqrt(((h2_pos-o_pos)**2).sum(axis=1)),
+    ])
+    if oh_dist.max() > 2.0: 
+        raise ValueError("Unphysical O-H bond length. Check that the molecules are entire, and atoms are listed in the expected OHH order.")
```
PicoCentauri: I would not constrain it in this way. The usual definition is as I did bonds should not be larger than half of shortest box length.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+    be "folded", and to be listed in the input as OHHOHHOHH."""
+    
+    o_pos = positions[::3]
+    h1_pos = positions[1::3]
+    h2_pos = positions[2::3]
+
+    oh_dist = torch.concatenate([
+            torch.sqrt(((h1_pos-o_pos)**2).sum(axis=1)),
+            torch.sqrt(((h2_pos-o_pos)**2).sum(axis=1)),
+    ])
```
PicoCentauri: I made quite some effort to not recompute distance but select the correct ones from the neighborlist...

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+    return oh_dist, hoh_angles
+
+def get_msites(system: System, m_gamma: torch.Tensor, m_charge: torch.Tensor):
+    
+    positions = system.positions
+    o_pos = positions[::3]
+    h1_pos = positions[1::3]
+    h2_pos = positions[2::3]
+    
+    # NB: this is enough to get the correct forces on the `system` atoms. thanks autodiff!
```
PicoCentauri: a bit too sloppy formulation for an example, but okay for now.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
-kinetic_energy = np.zeros(n_steps)
-total_energy = np.zeros(n_steps)
-temperature = np.zeros(n_steps)
-pressure = np.zeros(n_steps)
-trajectory = []
+    # adjust neighbor lists to point at the m sites rather than O atoms.
+    # this assumes this won't have atoms cross the cutoff, which is 
+    # of course only approximately true, so one should use a slighlty 
+    # larger-than-usual cutoff
+    # nb - this is reshaped to match the values in a NL tensorblock
```
PicoCentauri: Ahh true we have to update the NL... Thanks.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
-        atoms, rmax=6.0, nbins=200, elements=(8, 1)
-    )
-    HH_rdf_step, _ = ase.geometry.rdf.get_rdf(
-        atoms, rmax=6.0, nbins=200, elements=(1, 1)
-    )
-
-    OO_rdf_l.append(OO_rdf_step)
-    OH_rdf_l.append(OH_rdf_step)
-    HH_rdf_l.append(HH_rdf_step)
+class QTIP4PfModel(torch.nn.Module):
```
PicoCentauri: @I liked the idea of having a general class for three and four point water models.

Also if this should be the one and only QTIP4P model there should be not parameters for the user to set. They are fixed by the definition of the model.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+                m_gamma: float,
+                m_charge: float,
+                oh_bond_eq: float,
+                oh_bond_d: float,
+                oh_bond_alpha: float,
+                hoh_angle_eq: float,
+                hoh_angle_k: float,
+                p3m_options: Optional[dict]=None,
+                dtype: Optional[torch.dtype]=None,
+                device: Optional[torch.device]=None,
```
PicoCentauri: This should not be necessary anymore (See https://github.com/lab-cosmo/torch-pme/pull/166)

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+    ):
+        super().__init__()
+
+        if p3m_options is None:
+            # sane defaults, should give a ~1e-4 relative accuracy on forces 
+            p3m_options = (1.4, {'interpolation_nodes': 5, 'mesh_spacing': 1.33}, 0)
```
PicoCentauri: I would make them at least depend on the `cutoff`.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+        systems: list[System],
+        selected_atoms: Optional[Labels] = None,
+    ) -> tuple[System, TensorBlock]:
+        
+        """Remove possible ghost atoms."""
+        if len(systems) > 1:
+            raise ValueError(f"only one system supported, got {len(systems)}")
+
+        if selected_atoms is not None:
+            raise NotImplementedError("selected_atoms is not implemented")
```
PicoCentauri: You can use selected atoms: See the code I wrote.

examples/water-model/water-model.py
```diff
@@ -330,320 +354,269 @@
...
+        oo_distances = torch.linalg.vector_norm(neighbors.values[oo_mask], dim=1)
+        energy_lj = lennard_jones_pair(oo_distances, self.lj_sigma, self.lj_epsilon, self.cutoff).sum()
+        
+        # now this is the long-range part - computed over the M-site system
+        m_system, mh_dist, hh_dist = get_msites(system, self.m_gamma, self.m_charge)
+        potentials = self.p3m_calculator(m_system)
+
+        print("energies", e_bond, e_bend, energy_lj)
+
+        energy_tot = e_bond + e_bend + energy_lj
```
PicoCentauri: Coulomb?

PicoCentauri (commented):
Thanks for the update. I agree that including the class code in the example is a good idea.

However, I’m a bit disappointed that the effort I put into building three- and four-point models within a single class has been completely removed.

I also spent quite some time optimizing the code for efficiency, but these improvements are now lost because the implementation was rewritten from scratch in a more naive way. I’m okay with the new code, if the goal is to provide a clear and clean code (!) example that explains the background of how water models work, rather than focusing on performance.

examples/water-model/water-model.py
ceriottm: Given the new LAMMPS heuristics, I'd leave it as this - a O-H above 2Å means sure trouble

examples/water-model/water-model.py
ceriottm: If the molecules are ordered, this is likely to be far faster, but now the _auto branch uses distances everywhere - even to build M sites!

examples/water-model/water-model.py
ceriottm: I see your point here. I'll rename the class to TIP4P here and call the model qTIP4P/f. Re the 3-points, one can easily implement a 3-point by setting gamma to 0.

examples/water-model/water-model.py
ceriottm: hm. this I had found already like this. feel free to fix it before we merge if you think it can be avoided.

examples/water-model/water-model.py
ceriottm: then we should have heuristics in torch-pme. I don't think this is the place to introduce those heuristics

PR #110, opened 2025-01-22
Center cookbook text below logo

PicoCentauri:
I think looks better if the text below the logo is centered. I changed this by using a simple addition to the css

# Before

Screenshot 2025-01-22 at 13 40 42

# After

Screenshot 2025-01-22 at 13 40 26


ceriottm (approved):
I guess it's good to have an additional CSS in case we want to do more customization later.

PR #116, opened 2025-02-26
Add example to train a linear model for molecular polarizability 

ppegolo:
Draft for a recipe to train a linear ridge regression model with scikit-learn and featomic descriptors.
The model is then wrapped in a metatomic model and evaluated on ASE Atoms.


ppegolo:
@ceriottm were you thinking of something like this?

ceriottm (approved):
I did a few tweaks, now it looks good enough for a first "release". Maybe later (or now before merging) one could add a few more comments to the class explaining what it does, and use the calculator in a more ASE-y way (although I'm not sure we can...)

Issue #118, opened 2025-03-04
Zipped archive should contain all recipe files


PicoCentauri:
At the bottom of each recipe, we provide useful zipped archives. One for the code, one for the data files, and one for the environment file.

![Image](https://github.com/user-attachments/assets/6e901773-31bf-45e7-90b3-dae2784007df)

I think it would be useful to provide also a zipped archive that contains ALL of the necessary files together. There is already a button called `Download zipped`. However, this archive only contains the source code files. I think we should also add the data files and environment file to this zip archive.

pfebrer:
I also thought about this when running one of the examples.

I can try to solve this as a first issue 👍

ceriottm:
I agree it'd be much better. It's not entirely trivial I think because the zipped file is created by sphinx-gallery and it only includes the py and ipynb. The other files are created manually. Probably the best thing would be to disable the generation of this file by sphinx-gallery, and include the py and ipynb files with data in a single zip. I think there was a problem though. So this is just to say, thanks for looking into this but be careful, it's not entirely trivial

pfebrer:
Solved by #120

PR #120, opened 2025-03-04
Add full example into downloadable zip file.

pfebrer:
Addresses #118 by including all files of a given example in the downloadable zip file.

The approach that I took was to simply overwrite the zip file generated by `sphinx_gallery` with a zip file containing all the files. Then no extra download buttons are needed because `sphinx_gallery` already provides the download button for the zip file.

Some thoughts:

1. With the current implementation, the zip file will contain all the files that are present in the example directory. This might include outputs if this is not the first time that the build/tutorial is run. I think this is fine since in the CI the example directory will always be clean. Anyway, a `clean` command would also be nice to bring the directory to the original state, I don't know if that exists already.
2. I would say there is not really a need for the rest of download buttons since each individual download is not very useful, and having 5 stacked buttons looks ugly :)

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -290,12 +290,6 @@ for name in EXAMPLES:
         # Gather list of files before running the example
         example_files = list(example_dir.glob("*"))
 
-        # Creates data.zip if there's a data folder
-        if os.path.exists(f"examples/{name}/data"):
-            shutil.make_archive(
-                f"examples/{name}/data", "zip", f"examples/{name}/", "data/"
-            )
-
         session.run("python", "src/generate-gallery.py", example_dir)
 
         # Path of the generated gallery example.
@@ -389,10 +383,9 @@ that are not part of any of the other sections.
 """
         )
         # sort by title
-        for file, metadata in sorted(
+        for _, metadata in sorted(
             all_examples_rst.items(), key=(lambda kw: kw[1]["title"])
         ):
-            root = os.path.dirname(file)
 
             # generates a thumbnail link
             output.write(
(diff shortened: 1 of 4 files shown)


pfebrer:
I just found that there are some examples where the name of the python script is not the same as the example name. E.g. the `batch-cp2k` example has a `reference-trajectory.py` script and therefore the zip file is called `reference-trajectory.zip`.

Wouldn't it be better that the script was named `batch-cp2k.py`?

ceriottm:
> I just found that there are some examples where the name of the python script is not the same as the example name. E.g. the `batch-cp2k` example has a `reference-trajectory.py` script and therefore the zip file is called `reference-trajectory.zip`.
> 
> Wouldn't it be better that the script was named `batch-cp2k.py`?

sphinx-gallery allows having several `.py` files per folder, each of which is compiled to a separate page. I don't think we use this (yet), but I don't see why we should restrict examples to match the containing folder.

pfebrer:
Hmm yes but the way things are structured in `atomistic-cookbook` and what the code does both gave me the impression that your intentions are `one example == one script`. 

In any case the modifications that I did don't work for the case where there are multiple .py files (which create multiple gallery items and multiple downloadable zips), should I change it to support that?

ceriottm:
> Hmm yes but the way things are structured in `atomistic-cookbook` and what the code does both gave me the impression that your intentions are `one example == one script`.
> 
> In any case the modifications that I did don't work for the case where there are multiple .py files (which create multiple gallery items and multiple downloadable zips), should I change it to support that?

If it's not too hard, yes. You could basically include only the current .py and .ipynb file, and everything else. Should be just a regex. Note I'm also in the process of updating the instruiction to install, I'll push to this branch

pfebrer:
I think it's not too hard but it would complicate the code unnecessarily given that there is no need for it right now. And I don't know if other parts of the build would fail once you introduced the multiple .ipynb files. I can add a simple check for whether there are multiple .ipynb files and raise a meaningful error if that's the case. Like: "Hey, if you want multiple ipynb you need to implement this!"

pfebrer:
And also I don't know how we would use the multiple ipynb files exactly so I don't know which zipping would need to be implemented.

pfebrer:
> You could basically include only the current .py and .ipynb file

The thing is that there are no "current files", because the multiple files would be internally handled by `sphinx_gallery`, I think.

noxfile.py
ceriottm, line 1: Nice, you actually removed the ugly RST injection we had to do!

src/generate-gallery.py
```diff
@@ -70,6 +71,19 @@ def add_css_file(self, path):
...
+    # sure that the hack is still valid and it does not fail silently.
+    assert hasattr(sphinx_gallery.gen_rst, "ZIP_DOWNLOAD")
+    assert isinstance(sphinx_gallery.gen_rst.ZIP_DOWNLOAD, str)
+
+    sphinx_gallery.gen_rst.ZIP_DOWNLOAD = """
+    .. container:: sphx-glr-download sphx-glr-download-zip
+
+        :download:`Download recipe: {0} <{0}>`
+    """
+
```
ceriottm, line 86: Smart workaround!

ceriottm (approved):
Looks great, thanks for the contribution!

pfebrer:
Thanks for merging!

PR #122, opened 2025-03-12
An example of using the PET-MAD MLIP

ceriottm:


ceriottm:
Todo: 
- [x] Fix LAMMPS
- [x] Add a PET-MAD description
- [x] Move the mat-test xyz to a repository and fetch it dynamically
- [x] Add arxiv link and merge
(Add a simple UQ example) <-- this is postponed until we have a ckpt model with UQ

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,529 @@
...
+          Arslan Mazitov `@abmazitov <https://github.com/abmazitov>`_
+
+
+This example demonstrates how to use the PET-MAD model with ASE,
+`i-PI <https://ipi-code.org>`_ and `LAMMPS <https://lammps.org>`_.
+PET-MAD is a "universal" machine-learning forcefield trained on
+a dataset that aims to incorporate a very high degree of
+structural diversity.
+
+The point-edge transformer is an unconstrained architecture
```
frostedoyster:
```suggestion
The point-edge transformer (PET) is an unconstrained architecture
```

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,529 @@
...
+for more details). The unconstrained nature of the model simplifies
+its implementation and structure, making it computationally efficient
+and very expressive.
+
+The MAD dataset combines "stable" inorganic structures from the
+MC3D dataset, 2D structures from the MC2D dataset, and molecular
+crystals from the ShiftML dataset with "Maximum Atomic Diversity"
+configurations, generated by distorting the composition and structure
+of these stable templates. By doing so, PET-MAD achieves state-of-the-art
+accuracy despite that the MAD dataset contains fewer than 100k
```
frostedoyster:
```suggestion
accuracy despite the MAD dataset containing fewer than 100k
```

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,529 @@
...
+#
+# To start using PET-MAD, we first have to load the model. We will use the latest
+# version of the model. PET-MAD is distributed as a check-point file, that
+# allows re-training and fine-tuning, and needs to be *exported* to be used in
+# calculators and to make inference.
+
+mad_huggingface = (
+    "https://huggingface.co/lab-cosmo/pet-mad/resolve/main/models/pet-mad-latest.ckpt"
+)
+model = load_model(mad_huggingface).export()
```
frostedoyster: We need to be careful: this doesn't torchscript, potentially sacrificing some speed in ASE, which is what people will look for immediately.

The loss in speed can be avoided by saving and reloading, in two ways:
- either `model.save("model.pt", collect_extensions="extensions")`
- or `mtt export https://huggingface.co/lab-cosmo/pet-mad/resolve/main/models/pet-mad-latest.ckpt` from the command line

Then the ASE calculator below needs to be initialized with
`calculator = MetatensorCalculator("model.pt", device="cpu", extensions_dir="extensions")`

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,529 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._traj_atoms = []
+        self._traj_energy = []
+
+    def step(self, *args, **kwargs):
+        super().step(*args, **kwargs)  # Call original step function
+
+        self._traj_atoms.append(self.atoms.copy())
+        self._traj_energy.append(self.atoms.get_potential_energy())
```
frostedoyster: This is not the "ASE way" to do this, IMO. I would `attach()` functions to the LBFGS object

frostedoyster (commented):
I think that this example trying to do too much. Simplifying it and/or providing a split version in the new PET-MAD repo are possibilities to make it more accessible

examples/pet-mad/pet-mad.py
ceriottm: Feel free to change this to whatever you feel is ASE-esque enough.

examples/pet-mad/pet-mad.py
ceriottm: Right, makes sense.

ceriottm: OK addressed this and it is also a better point to discuss exporting. I also explained a bit how ckpt and pt differ.

ceriottm:
> I think that this example trying to do too much. Simplifying it and/or providing a split version in the new PET-MAD repo are possibilities to make it more accessible

I disagree. It is a lot, but it does showcase the accuracy and usage of pet-mad in most of the plausible use cases. I also think that once we have a .ckpt with LLPR it'd be nice to show how to use it here. We should and will make more separate examples (ensemble propagation, fine-tuning, etc.) but having a one-stop showcase of the accuracy of PET-MAD, and of what it can do to model complicated systems with the three modeling codes it currently supports, is better than having people download 3 different mini-recipes.

examples/pet-mad/pet-mad.py
frostedoyster: Done

Luthaf:
Regarding having a single example that does everything or multiple smaller and focused examples; one more point in favor of multiple example is that it allow us to keep reminding people about PET-MAD, by making one announcement + marketing push for each example as we release them over a couple of week. Such repeated exposure would make sure that everyone is aware of PET-MAD and that it can do a lot of things, and would be a lot more effective that dumping everything at once and letting people realize wide variety of use cases by themself.

EDIT: overall I think this example is not too long by itself, but it should not contain much more than it currently does IMO.

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,541 @@
...
+    get_ipython().run_line_magic("matplotlib", "inline")  # noqa: F821
+
+
+# %%
+# Inference on the MAD test set
+# -----------------------------
+#
+# We begin by using the ``ase``-compatible calculator to evaluate energy and forces for
+# a test dataset that contains both hold-out structures from the MAD dataset, and a few
+# structures from popular datasets (MPtrj, Alexandria, OC2020, SPICE, MD22) re-computed
```
Luthaf: should this contains links to the datasets?

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,541 @@
...
+model = load_model(mad_huggingface).export()
+
+# %%
+#
+# The model can also be downloaded separately, and loaded from disk by providing the
+# path to the :py:func:`load_model <metatensor.utils.io.load_model>` function.
+#
+# This model can be used "as is" in Python - and in this form one can modify it, e.g. to
+# continue training, or to fine-tune on a new dataset. However, to run with external
+# codes, it can/should be saved to disk.
```
Luthaf:
```suggestion
# codes such as LAMMPS, it can/should be saved to disk.
```

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,541 @@
...
+# without requiring the original Python or C++ source code. In particular,
+# this is necessary for the LAMMPS interface to work because it has no access to
+# the Python code.
+
+# %%
+#
+# Single point energy and forces
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# PET-MAD is compatible with the metatomic interface which allows us to
```
Luthaf:
```suggestion
# PET-MAD is compatible with the metatensor atomistic models interface which allows us to
```

metatomic does not exist yet, this could confuse people

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,541 @@
...
+#
+# Here, we run the computation on the CPU. If you have a CUDA GPU you can also set
+# ``device="cuda"`` to speed up the computation.
+
+mad_energy = []
+mad_forces = []
+mad_structures = []
+for structure in test_structures:
+    tmp = deepcopy(structure)
+    tmp.calc = copy(calculator)  # avoids ovewriting results. thanks ase 3.23!
```
Luthaf:
```suggestion
    tmp.calc = copy(calculator)  # avoids ovewriting results.
```

examples/pet-mad/pet-mad.py
```diff
@@ -0,0 +1,541 @@
...
+    settings=chemiscope.quick_settings(
+        x="energy_mad",
+        y="energy_ref",
+        symbol="origin",
+        structure_settings={"unitCell": True, "shape": ["forces_ref"]},
+    ),
+)
+
+# %%
+# How about equivariance!?!1?1!
```
Luthaf:
```suggestion
# How about equivariance‽
```

examples/pet-mad/pet-mad.py
PicoCentauri: I am sure, I fixed this in my PR a couple of days ago...

examples/pet-mad/pet-mad.py
PicoCentauri: Same here...

examples/pet-mad/pet-mad.py
ceriottm: Owwwww, you guys are far too serious.

examples/pet-mad/pet-mad.py
ceriottm: added!

ceriottm:
> EDIT: overall I think this example is not too long by itself, but it should not contain much more than it currently does IMO.

Agreed. The idea is to then have several focused examples - one on ensemble propagation, one on fine tuning, etc.

PR #123, opened 2025-03-19
Fix link in pet-mad recipe

PicoCentauri:


ceriottm (approved):
there are times where bypassing branch protection is really the right thing to do

PR #124, opened 2025-04-10
A notebook for non-conservative MTS dynamics with PET-MAD

ceriottm:
A recipe to run MD using non-conservative forces from PET-MAD, using MTS to avoid artefacts


examples/pet-mad-nc/pet-mad-nc.py
```diff
@@ -0,0 +1,352 @@
...
+
+import ase.io
+
+# i-PI scripting utilities
+import chemiscope
+import matplotlib.pyplot as plt
+import metatensor.torch.atomistic as mta
+
+# import numpy as np
+# import requests
```
frostedoyster: To be removed?

frostedoyster (commented):
Are the Ni and Al structures still used?

The recipe looks good

frostedoyster:
We could also set the verbosity level to "low". At the moment, a lot of redundant i-PI information is being printed in the recipe

ceriottm:
> Are the Ni and Al structures still used?
> 
> The recipe looks good

No as far as I can tell they both have been removed. I'm using the BMIM-Cl example that is a very challenging one, and so gives large energy drift in a short time.

PR #126, opened 2025-04-24
Example using the MCoV model

ppegolo:
Example showcasing the MCoV model for equivariant predictions based on scalar features using a tensorial basis.


ceriottm:
I'm looking at the recipe - there has been some copy-pasting from the linear model recipe that need to be checked - e.g. the description of the dataset is wrong for both, in the linear model it seems to have sulfur, and this is not a single composition. I'll also keep going through it but @ppegolo it'd be good if you could re-read both with an eye to the detail.

ppegolo:
> I'm looking at the recipe - there has been some copy-pasting from the linear model recipe that need to be checked - e.g. the description of the dataset is wrong for both, in the linear model it seems to have sulfur, and this is not a single composition. I'll also keep going through it but @ppegolo it'd be good if you could re-read both with an eye to the detail.

Oops, forgot to change that line. I'll check them

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+
+# %%
+# The :math:`\lambda`-MCoV model
+# ------------------------------
+#
+# Here is a schematic representation of the :math:`\lambda`-MCoV model which, in a
+# nutshell, allows us to learn a tensorial property of a system from a set of scalar
+# features used as linear expansion coefficients of a minimal set of basis tensors.
+
+fig, ax = plt.subplots(figsize=(5728 / 300, 2598 / 300), dpi=300)
```
ceriottm, line 179: You can import the figure without doing all this plt magic. See pi-mts-rpc example, you basically just use a RST block

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+
+# %%
+#
+# We parametrize spherical tensors of order :math:`\lambda` as linear combinations
+# of a small, fixed set of **maximally coupled** basis tensors. Each basis tensor is
+# computed from three learned **vector** features, and each coefficient is predicted by
+# a **scalar** function of the local atomic environment. This enforces exact
+# equivariance under the action of the orthogonal group O(3), while relying only on
+# efficient scalar networks.
+#
```
ceriottm: Remove whiteline, otherwise it gets formatted as a box. 
```suggestion
```

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+#   (\mathbf{q}_1,\mathbf{q}_2,\mathbf{q}_3) + \mathbf{T}_\lambda^{\mathrm{corr}}
+#   (\{\mathbf{r}_i\}),
+#
+# where :math:`\mathbf{B}^\lambda_\beta` is a shorthand for the basis tensors and
+# :math:`s_\beta` for the scalar coefficients. For global properties we sum over all
+# atoms.
+
+# %%
+# Training and evaluation of the model
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ceriottm:
```suggestion
# ------------------------------------
```
This should be a top-level title

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+#
+# We train the model with the ``metatrain`` package via its command-line interface.
```
ceriottm:
```suggestion
# Rather than implementing the :math:`\lambda`-MCoV model from scratch, 
# we use a pre-defined architecture within
# the ``metatrain`` package, using its command-line interface.
```

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+# To start training, we run
+#
+# .. code-block:: shell
+#    mtt train data/options.yaml
+#
+# The options file specifies the model architecture and the training parameters:
+#
+# .. literalinclude:: data/options.yaml
```
ceriottm: This is not working, you should check why. I suspect that stuff in data/ is not copied over when the sphinx-gallery files are moved to the docs/examples folder.

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+#    :language: yaml
+#
+
+# %%
```
ceriottm, line 332:
```suggestion
# %%
# To execute ``metatrain`` from within a script, use
```

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+plt.loglog(train_log["Epoch"], train_log["validation_loss"], label="Validation")
+plt.xlabel("Epoch")
+plt.ylabel("Loss")
+plt.legend()
+plt.title("Training and validation loss")
+
+# %%
+# We evaluate the model on the test set using the ``metatrain`` command-line interface:
+#
+# .. code-block:: shell
```
ceriottm, line 365: You need an empty line to start a block in RST - this is why all (most) the code blocks are not show in the rendered page.

ceriottm: Note that when you build offline there should be warnings that you can use to hunt these down.

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
```diff
@@ -0,0 +1,445 @@
...
+#    mtt eval model.pt eval.yaml -e extensions -o test_results.mts
+#
+# The evaluation YAML file contains lists the structures and corresponding reference
+# quantities for the evaluation:
+#
+# .. literalinclude:: eval.yaml
```
ceriottm, line 372: Missing, as above

ceriottm (changes requested):

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
ppegolo: I did that to be able to use the image as thumbnail

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
ppegolo: `.yaml->.yml` made it

ceriottm: metatrain expects .yaml though, no? you can also (and maybe it's better) modify 
`src/generate-gallery.py:            "copyfile_regex": r".*\.(sh|xyz|cp2k|yml|png|zip)"`
to add yaml as one of the files that get copied over

examples/learn-tensors-with-mcov/learn-tensors-with-mcov.py
ceriottm: uhm uhm uhm. it's a lot of code might be a bit confusing. add at least a comment to say this is to generate a thumbnail
```suggestion

# visualize a scheme of the architecture that is also used as a thumbnail for the recipe
fig, ax = plt.subplots(figsize=(5728 / 300, 2598 / 300), dpi=300)
```

ceriottm (approved):
All looks great now

PR #127, opened 2025-04-28
Update environment.yml

ppegolo:


Luthaf (commented):
I don't think we should use strict version pin `foo ==1.2.3` but rather use semver pining `foo >=1.2.3,<2.0`. Otherwise people trying to use these example will install versions of the different packages that contains fixed bugs.

ceriottm:
I'm going to fix this this morning, promised.

ceriottm:
I suggest that for cleanliness we close this branch, and start a new one in which we agree upon, implement and document a common policy for version pinning. I think we need to find a balance between keeping examples up-to-date and asking people making a new example have to worry about updating all the examples that broke in the meantime.

PR #129, opened 2025-04-29
Update PET-MAD example to PET-MAD v1.1.0rc4

abmazitov:


examples/pet-mad/environment.yml
```diff
@@ -3,10 +3,8 @@ channels:
   - conda-forge
 dependencies:
   - python=3.12
-  - lammps-metatensor=*=cpu*nompi*
-  - libtorch=*=cpu*mkl* 
-  - pet-mad
-  - python-metatensor-torch=0.7.6  
+  - lammps-metatensor=2025.02.04.mts0=cpu_hb5fed64_nompi_git.ce972db_0
+  - pet-mad=1.1.0rc4=pyhb0d71a2_0
```
Luthaf: Please do not pin to specific versions, but use `>=` and `<` where relevant

```suggestion
  - lammps-metatensor
  - pet-mad >=1.1.0rc4
```

abmazitov: Why not? It was @ceriottm who asked me to pin the exact version

Luthaf: Because if everyone pins everything everywhere, it gets very painful. We could introduce lockfiles to save the exact versions of a set of dependencies known to work together in the cookbook, but the current situation where half of the dependencies are pinned and the other half is not means a lot of churn and repeat work whenever we do an update.

Luthaf: As mentioned in https://github.com/lab-cosmo/atomistic-cookbook/pull/127#pullrequestreview-2802854292, we also don't want people who are using these example to be stuck on old versions, with bugs already fixed in the latest releases.

IMO the default behavior should be that all examples run on the latest version of everything, and we can deviate on a case by case basis if there is a good reason

abmazitov: Well, okay, that works

ceriottm: we saw clearly this didn't work though. the pet-mad example broke, and then we couldn't really merge anything new until that was fixed. I think that expecting to keep the whole thing running, and expecting people to fix other examples whenever they want to update one is not realistic.

ceriottm: I think we could use semver pinning for stable packages, but pin exactly metaverse packages that are still changing too quickly

ceriottm:
OK, merging this and we move on the discussion on pinning to #127

ceriottm (approved):

Issue #130, opened 2025-04-30
Run only touched examples in pull requests CI?


PicoCentauri:
We currently face the issue that if one recipe is broken, the whole CI workflow is blocked. This is especially a blocker if another recipe is updated. The broken recipe is blocking the CI and makes reviewing and proofreading the rendering inside a PR very hard.

I suggest that inside a pull request, we only check touched recipes. Once a PR is merged into the main branch, we should for sure run all recipes to create the new page.

pfebrer:
I think this makes sense, but what if for example you want to check if links to other recipes work? (I don't know if creating such links is even possible though). Should we have a cache with the latest build of each recipe? Then we can generate the full website even if the build of one recipe is currently failing

pfebrer:
(I can take care of this once we agree on what's the best thing to do)

ceriottm:
My main concern here is that (1) we want a website that keeps building correctly and (2) we don't want to showcase broken recipes. To me if we want to only check the touched recipes, we should also cache the built pages so that we don't break the website. Maybe the way to go is to implement a sane pinning policy (see #131) and then have a CI job that is run once a week and notices the documentation team whenever one of the examples stops building.

Luthaf:
> I suggest that inside a pull request, we only check touched recipes. Once a PR is merged into the main branch, we should for sure run all recipes to create the new page.

This would mean that we only realize that the website is broken AFTER merging a PR, which I really don't want.

> we should also cache the built pages 

This would be nice, but I can't think of an easy way to do this. If we store build artifact somewhere, we need quite a bit of code to update/invalidate the cache, and we need to integrate with some caching service somewhere. Alternatively, if we commit all generated files to this repo, the size of it will explode immediately (since there are quite a few examples that download/generate large files).

----

An alternative solution would be to have a list of `broken` examples which would be skipped by CI altogether until someone comes and fixes it. We could automate opening issues when a new example is added to this list and pinging the authors so they are aware of the breakage.

pfebrer:
Yeah I think caching the builds is something that we must do, together with any other measure. It makes no sense that if someone wants to contribute they have to fix the broken recipes first.

Caching could be as simple as keeping a `build` branch on this repo with the latest builds.

pfebrer:
Ok I can see how big files could be a problem if there are any, but still think some caching implementation is worth the effort.

pfebrer:
[...]

PR #131, opened 2025-04-30
Apply a coherent version pinning policy

ceriottm:
We need to agree on a strategy to have examples run stably in the future, as it is already too much a pain to make a new recipe if you also have to worry about updating all other recipes. This PR is to agree on a policy and apply it to all current recipe environments.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -58,7 +58,7 @@ this, e.g. indicating the version known to work and the next major update
 
    - my-package >=1.3.1,<2.0
 
-This way if the example breaks further down the line it is easy to 
+This way if the example breaks further down the line it is easy to
 bring it back to a working state converting these to strict pinnings,
 yet most harmless updates will be applied automatically.
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -49,7 +49,7 @@ Package pinning
 The Python ecosystem moves fast and breaks things. Examples should always
 be able to run in the CI system to generate webpages, and it's best for
 users if they also work out-of-the-box when downloaded. For these reasons,
-we recommend pinning the package version in the environment file.
+we recommend constraining the package version in the environment file.
 At the same time, it's best to run on updated versions that might contain
 important bugfixes. We recommend using semantic versioning to achieve
 this, e.g. indicating the version known to work and the next major update
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -43,8 +43,8 @@ following files inside:
 - as many Python files as you want, each one will be converted to a separate
   HTML page.
 
-Package pinning
-~~~~~~~~~~~~~~~
+Example dependencies
+~~~~~~~~~~~~~~~~~~~~
 
 The Python ecosystem moves fast and breaks things. Examples should always
 be able to run in the CI system to generate webpages, and it's best for
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -43,6 +43,28 @@ following files inside:
 - as many Python files as you want, each one will be converted to a separate
   HTML page.
 
+Package pinning
+~~~~~~~~~~~~~~~
+
+The Python ecosystem moves fast and breaks things. Examples should always
+be able to run in the CI system to generate webpages, and it's best for
+users if they also work out-of-the-box when downloaded. For these reasons,
+we recommend pinning the package version in the environment file.
+At the same time, it's best to run on updated versions that might contain
+important bugfixes. We recommend using semantic versioning to achieve
+this, e.g. indicating the version known to work and the next major update
+
+.. code-block:: yaml
+
+   - my-package >=1.3.1,<2.0
+
+This way if the example breaks further down the line it is easy to 
+bring it back to a working state converting these to strict pinnings,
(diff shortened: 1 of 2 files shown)


ceriottm:
Picking up the slack from #127

pfebrer:
Whenever a recipe runs succesfully, we could create a yaml file with the exact environment that produced the build. Then it would be easier to debug, and using https://github.com/lab-cosmo/atomistic-cookbook/issues/130 we could delay debugging to whenever we have time to do it.

As a safeguard, we could provide this fixed environment yaml for the user to download, then we make sure that the user has at least one way to run the recipe without problems (if the normal specification fails).

I don't think pinning the dependencies on all examples is a good idea, since we would have to constantly make sure that the pin on all examples is set to the maximum possible, and dealing with multiple pinned dependencies is a nightmare :sweat_smile:

Luthaf:
What @pfebrer is suggesting is often called a lockfile, and I agree we could do this, although it would be a bit of work to generate them & enable using them.

An alternative as well would be to make it easier to deactivate specific example, this way if something unrelated breaks when working with this repo, the corresponding example can be deactivated and fixed later instead of blocking everyone else. IMO this would give most of the benefits of pinning for contributors to this repo with none of the drawbacks.

--- 

My suggestion is to make pinning the exception rather than the rule. We can pin things when:
- the package API is very unstable and version number does not use semver. Out of the metaverse, I think this is only the case for metatrain
- there was a breaking change we don't have the time to deal with right now

In both cases, the pin should come with a comment explaining why it is necessary, so the next person touching it knows what to check.

CONTRIBUTING.rst
```diff
@@ -43,6 +43,28 @@ following files inside:
 - as many Python files as you want, each one will be converted to a separate
   HTML page.
 
+Package pinning
+~~~~~~~~~~~~~~~
+
+The Python ecosystem moves fast and breaks things. Examples should always
+be able to run in the CI system to generate webpages, and it's best for
+users if they also work out-of-the-box when downloaded. For these reasons,
+we recommend pinning the package version in the environment file.
```
Luthaf:
```suggestion
users if they also work out-of-the-box when downloaded. For these reasons,
we recommend constraining the package version in the environment file.
```

for me a pin is `foo ==1.2.3`

CONTRIBUTING.rst
```diff
@@ -43,6 +43,28 @@ following files inside:
 - as many Python files as you want, each one will be converted to a separate
   HTML page.
 
+Package pinning
+~~~~~~~~~~~~~~~
```
Luthaf:
```suggestion
Example dependencies
~~~~~~~~~~~~~~~~~~~~
```

pfebrer:
> although it would be a bit of work to generate them & enable using them.

Why? This is just

```
conda env export > environment-lock.yaml
```

no?

Luthaf:
Does this account for `pip` dependencies inside conda?

ceriottm:
I think that having versions listed with >=known_working,<next_major is a good compromise, then it's easy to switch to a pin to get an obsolete example working

Luthaf:
The other issue with lockfiles is that a lockfile generated on one system (e.g. linux-x64) might not work on another (e.g. macOS or linux-arm64)

pfebrer:
> Does this account for pip dependencies inside conda?

Yes, it produces something like:

```yaml

dependencies:
    -  ...
    -  ...
    - pip:
        - ...
        - ...
        - ...
```

pfebrer:
> The other issue with lockfiles is that a lockfile generated on one system (e.g. linux-x64) might not work on another (e.g. macOS or linux-arm64)

Can't this be said about the normal specifications as well? There is no way to know unless you test on all platforms.

pfebrer:
> I think that having versions listed with >=known_working,<next_major is a good compromise, then it's easy to switch to a pin to get an obsolete example working

In the PET-MAD example you set it to  >=known_working,<next_minor no?

Luthaf:
> Can't this be said about the normal specifications as well? There is no way to know unless you test on all platforms.

Well, the normal specifications can be made compatible with multiple versions (like we do for lammps-metatensor in the pet-mad example). But you are right this is not a guarantee that the code will work =)

ceriottm:
IMO "recommendations" is the best thing for a project like this, without being too strict.
Let's leave contributors some freedom to do what makes sense for each recipe, and if anything provide tools to facilitate their work. I 100% agree that a simple way to "deprecate" examples would be important to have. This could be as simple as having a deprecated/ folder in the repo, possibly with statically stored RST from sphinx so we can still include them in the website.

Luthaf:
I was more thinking of a list of "broken" example in the nox configuration, so we can very easily disable broken example, without moving a bunch of files around. The examples would then be out of the website until someone fixes them (unless we find a way to store the last-working rst somewhere)

pfebrer:
I think this can be merged, no?

pfebrer:
I'll merge it this afternoon if there are no objections

Luthaf (approved):

pfebrer:
So I merged this and now it is failing (https://github.com/lab-cosmo/atomistic-cookbook/actions/runs/14837300428) I think because `ipi` was moved to 3.1.3 and it doesn't work (the dependency is specified like "ipi>=3.1.2,<3.2").

@ceriottm do we pin `ipi` then?

PR #132, opened 2025-04-30
On PR only run modified examples

### LINKED ISSUE #130: Run only touched examples in pull requests CI?

pfebrer:
This should work, I have tested it except for the github action, which I will test with a dummy PR.

Closes #130 by running only the modified examples during the CI of a PR. If a file is modified outside the `examples` directory everything is ran, since there is no way to tell which examples will get affected.


Luthaf:
I'd say if there are modifications to files outside of the examples, we should re-run all examples in CI. This allows for example to test modifications to nox, or the global sphinx configuration.

pfebrer:
Yes this is what will happen

src/get_examples.py
```diff
@@ -18,10 +26,62 @@ def get_examples():
...
+    """
+    modified_files = [file.resolve().relative_to(ROOT) for file in modified_files]
+
+    if any(
+        file.parts[0] != "examples" and str(file) not in IGNORE_CHANGES
+        for file in modified_files
+    ):
+        # This means that there is a modified file outside of the examples directory.
+        # In this case, all examples *could* be affected by the change.
+        return get_examples()
```
pfebrer, line 48: @Luthaf This is the case in which we just can't tell which examples will get affected.

Luthaf: thanks, I tried to find this but missed it!

src/get_examples.py
```diff
@@ -3,10 +3,18 @@
 import glob
 import json
 import os
+from pathlib import Path
 
 
 ROOT = os.path.realpath(os.path.join(os.path.dirname(__file__), ".."))
 EXAMPLES = os.path.join(ROOT, "examples")
+IGNORE_CHANGES = [
```
ceriottm, line 12: These get incorporated into the docs, but only if the `build_docs` is done. Is this the case?

src/get_examples.py
```diff
@@ -18,10 +26,63 @@ def get_examples():
...
+        return (
+            file.parts[0] == "examples" and Path(file.parts[0], file.parts[1]).is_dir()
+        )
+
+    return list(
+        set(file.parts[1] for file in modified_files if is_inside_example(file))
+    )
+
+
+def create_json(modified_files: str):
```
ceriottm: maybe a more telling name would be `create_example_ci_jobs`?

ceriottm (approved):
Ok just a couple of small comments, feel free to merge after you have considered/addresed them.

src/get_examples.py
pfebrer: I don't know what you mean exactly. What I did here is to avoid that changes to these files trigger CI builds of all the examples, if that answers the question 😅

pfebrer: I will add a comment to the variable explaining how it is used

PR #133, opened 2025-05-07
Make pull request CI runs great again

pfebrer:
In this PR three things are modified:

1. Most files outside the `examples` directory don't affect the internal running of examples. Now, the files that do affect the examples are listed explicitly. They are the only ones that are checked for modifications to determine whether CI should be ran.
2. There might be cases in which there are no examples to build. In that case the CI was failing, so I added some code to handle the 0 examples case.
3. The examples that are not built in the CI run are retrieved from the latest successful build of the main branch. In this way the PR can build the full documentation without wasting CI time :heavy_check_mark:


pfebrer:
Surprisingly it works and implementing caching by reusing the artifacts was easier than I expected :)

You can put the implementation to the test by pushing changes to one example in this PR.

pfebrer:
Ok, so when the CI ends (I modified noxfile.py so everything will run) this is ready to merge. As a final touch I added the possibility to download the example artifacts locally. As documented in `CONTRIBUTING.rst`, this is done like:

```
python src/latest_docs_run.py download-examples
```

and the CLI has options to choose which examples to download. Then debugging CI locally should be easier.

PicoCentauri (approved):

PR #134, opened 2025-05-14
Finetuning

DavideTisi:


ceriottm:
Very nice. A few comments for improvement:
1. would be nice to show training curve (train and validation error vs epoch). also it's not clear if ethanol is computed with a different flavor of DFT, you should indicate that. 
2. It'd be nice (even if it make the model slower) demonstrate the accuracy of a model trained "from scratch" for 1000 epochs. the comparison between training curves would make a nice fig. 1
3. I'd also show training curves for the conservative fine-tuning, maybe showing the "appended" curve plot as in the paper
4. another nice thing would be to show in the final parity plot the last point in the NC pre-training in the background with low opacity, so it's clear how much it improved in just 50 epochs.

ceriottm (approved):
LGTM

PR #139, opened 2025-05-27
CI: Add an action to prevent large files from sneaking into the repo

HaoZeke:
Adds a CI stage to `lint` which checks for large files [1].

[1]: https://github.com/HaoZeke/large-file-auditor


pfebrer:
Do we need to check all the files? I thought we wanted to check only the ones getting pushed by the PR

HaoZeke:
> Do we need to check all the files? I thought we wanted to check only the ones getting pushed by the PR

Easier to parse the history out of the entire repo, the issue is that we don't want just the changed files (reworking the action to point to commits instead of the blob IDs). We can run `du` on the files changed in a PR, but that would be hard(er) to pin down than filtering with `git filter-branch`. Plus its very quick as-is.

pfebrer:
Yes but filtering the whole history is something to do just once in principle, and when someone wants to merge a PR they should get feedback only on what they are merging. They should not get their PR blocked by something that has been done wrong previously.

HaoZeke:
> Yes but filtering the whole history is something to do just once in principle, and when someone wants to merge a PR they should get feedback only on what they are merging. They should not get their PR blocked by something that has been done wrong previously.

Although if they are changing say, the file size, they'll necessarily have to get all the commits audited. In general when its green its green, and if there is a large file somewhere in history it needs to be fixed sooner rather than later right? The idea is that this CI run should always be green... ~as it is now.~ Acutally @Luthaf mentioned 7M is way too generous so we'll need another look.

HaoZeke:
OK this is now ready for review, basically needs a decision on what size we'd like to limit to.

cc @DavideTisi @pfebrer 

@ceriottm suggested 1M (and that we should optimize the images anyway).
@Luthaf prefers a much smaller limit, around 100K or 200K (which means the images all need to be reworked)

I'm pretty ambivalent, the general rule is that it should be low, and also that it shouldn't be modified much, i.e. 1M is OK if everything is basically only touched once.


EDIT: Ready for review but still needs a cleanup and force push, the current failing commits are::

```bash
# examples/torchpme/data/point_charges_Training_set.xyz
➜ git log --all --pretty=format:%H --find-object=c5e3cf0a8c33810e3c3b26927f7e4b35ae8708f1
46ba137be93733e5ee8d55d019303f696de67474

# examples/sample-selection/data/input-fps.xyz
atomistic-cookbook on  testFilterAction via 🐍 v3.13.3 
➜ git log --all --pretty=format:%H --find-object=94414ecacce998c3b845c27d7cb22c3023a8978d                                       
4107b33703bc3876cded6f5bff2be1d4c4782c4d
78de4db6b3b97a61e00adf8154a0f0331c339d3b
1e42c8f1aeaf50260e820294f87f0f3a0e5348e0
2ad2cae8da3cff05431824f0ab04d6db9a73b4b9
```

Where the problem is evident, the smaller file is in a bunch of commits. Easily removed with https://github.com/newren/git-filter-repo still, but something to keep in mind.

pfebrer:
Given that this is a repo of tutorials I think small images with a couple hundred KB are fine. It should be as easy as possible for people to contribute, and if the limit is lowered too much it can become annoying.

pfebrer:
Also, perhaps the external action should go to `lab-cosmo` instead of being hosted in a personal GitHub profile?

pfebrer:
Or it could be simply a script so that it can also be ran locally?

DavideTisi:
I agree with @pfebrer, I would limit the limitations to not make contributing more annoying

HaoZeke:
I could move the action under `lab-cosmo`, wasn't sure it fit in with the rest of the org though (typescript GHA). More of a general utility.

As to making a script, yep we could implement it in one block here, but that's more fragile than using an existing component IMO.

ceriottm:
One important thing: this should not prevent all other tests from running, we want ppl to be able to develop with data in the folder, and only create an archive when they're ready to merge

HaoZeke:
So then 1M sounds good to everyone, and someone needs to rework the commits mentioned above maybe just with `git filter-repo` via:


```bash
pipx run git-filter-repo --invert-paths \
  --path 'examples/torchpme/data/point_charges_Training_set.xyz' \
  --path 'examples/sample-selection/data/input-fps.xyz'
```

and a force push.


If we really want to cleanup the repo size, there are more files to filter out from the history and logs, e.g. 

```bash
pipx run git-filter-repo --analyze
cat .git/filter-repo/analysis/blob-shas-and-paths.txt | head
=== Files by sha and associated pathnames in reverse size ===
Format: sha, unpacked size, packed size, filename(s) object stored as
  78e85678d3dda8725d3a9a2bd112ee96e246f808   60925256   55928633 examples/hamiltonian-qm7/data/qm7/def2-tzvp/fock.hickle
  116e885a3be9eb4f64dc9a0c9adafa4562cadc80   60925256   45454903 examples/hamiltonian-qm7/data/qm7/def2-tzvp/overlap.hickle
  23b0eeb1fd3d5bce4634ce38eb851f76f4f74ca5   23339848   23346187 _datasets/40e1de84c3a8b7e0a3887dc1dcce5e69c044097c682fbd2cd11a7bfd82f0cc15-fig_gaas-map_001.json.gz
  ee339b844f66dcf44ac0cf833bc2f80f84a42eba   22912056   12981055 examples/hamiltonian-qm7/data/ethane/def2-tzvp/fock.hickle
  29cff519317a3beaa2df109f554d6526a8e93527   10817492   10328523 examples/ipi-nanopet/inputs/model-zundel.pt
  4196fbae9dbe8a002a5608ac6b4613a891297386   22912056    9707160 examples/hamiltonian-qm7/data/ethane/def2-tzvp/overlap.hickle
  8f69fab797cd21322fddad2d10228c1f1b0d7614    6939718    4428158 _sources/examples/periodic-hamiltonian/periodic-hamiltonian.rst.txt
  fdd0ff96ac997efda6f5d745fda8d598e898cd69    6608110    4365856 examples/periodic-hamiltonian/periodic-hamiltonian.html
```

Where the `hickle` file shows up in:

```bash
➜ git log --all --find-object=78e85678d3dda8725d3a9a2bd112ee96e246f808 --pretty=%H
0485322b044946d8235dd36d3cc969c86a6f0eeb
6bad34c29a4a1c4bae4ba698956797211f9704c9
```

Deletion is still with `git filter-repo` and the path, but each operation is basically a history rewrite so I'd suggest doing it only once...

HaoZeke:
> One important thing: this should not prevent all other tests from running, we want ppl to be able to develop with data in the folder, and only create an archive when they're ready to merge

The `lint` job isn't a blocking one, so it shouldn't matter if it fails a bunch while people push things. Just before a squash merge amending the commit should work.

In general since we're talking best practices we might as well explicitly add `*.xyz` and `*.tar.*` to the `.gitignore` so they have even less chances of sneaking in.

pfebrer:
The `.hickle` files show up because of the current `hamiltonian-qm7` branch, but Hanna already pushed the necessary changes and they will disappear tomorrow when we squash commits.

I think it's better not to `.gitignore` things. Again it can make it unnecessarily annoying for people to contribute, and a < ~100KB xyz file I would say it is fine for the tutorials.

pfebrer:
So I guess we need @E-Rum or @PicoCentauri to provide a solution for the `examples/torchpme/data/point_charges_Training_set.xyz` file, and @DavideTisi or @HannaTuerk to provide a solution for the `examples/sample-selection/data/input-fps.xyz` file. Sorry people :smile:

DavideTisi:
So is it the max 1M? becouse in that case, for FPS, the solution is simply to cut the dataset in half

ceriottm:
> So is it the max 1M? becouse in that case, for FPS, the solution is simply to cut the dataset in half

I think the idea is to offload to a repository as much as possible. In this case, the simple solution might be to pick a dataset that is openly available  and can be easily downloaded - the GaAs would be a possibility (already used in another recipe)

HaoZeke:
For this kind of smallish total data (Dagshub allows 20GB total), I'd actually recommend DVC [1] and DagsHub [2], it is another tool to keep track of but it's basically `git` for data, it'd leave `.dvc` files in the repo which are refs to the objects stored on DagsHub. Much nicer than handling Zenodo / Figshare etc with each change.

[1]: https://dvc.org/
[2]: https://dagshub.com/

ceriottm:
@E-Rum , I presume that the point-charge dataset is available somewhere. could you please change the charge example so it downloads it from there rather than have it bundled, so we can merge & close this PR?

ceriottm (approved):
CI action works (small problem is that it's not tested locally, but doesn't really matter IMO) and the two recipes that were breaking the limit have been sanitized. Merging now!

PR #141, opened 2025-05-28
Hamiltonian qm7 - now without large dataset

HannaTuerk:
This is an example of indirect learning of the Hamiltonian in the spirit of [arXiv:2504.01187](https://arxiv.org/abs/2504.01187). Datasets used are of a rattled ethane and a small set of qm7 molecules is used. The dataset is now downloaded from zenodo.


pfebrer (approved):

pfebrer:
Ok, the hickle files are no longer in history, but I just saw that there is a best-model.pt file which is 1.2MB, so according to the new rules it will have to get removed as well :cry:

HannaTuerk:
noooooooo
I guess I can make a new zenodo that includes this file.. to be continued..

PR #142, opened 2025-05-28
Hamiltonian qm7 - now dataset and model are on Zenodo

HannaTuerk:
Neverending Story - Chapter 3
Example of indirect Hamiltonian learning written by @DivyaSuman14 and me in the spirit of [arXiv:2504.01187](https://arxiv.org/abs/2504.01187)  for small sub-selected datasets of rattled ethane and QM7. Dataset and model  are now published on Zenodo. No large files are present, let's hope this time we do not have to rewrite history.


pfebrer:
@DivyaSuman14 do you know how much time would it take to produce the plots for the best model instead of having the images saved?

DivyaSuman14:
In theory, generating the plots directly doesn't take much time. However, due to the current structure of our code, the process is a bit more complex. To load a saved model, we first need to initialize it. Unfortunately, our initialization routine depends on precomputed features because one model needs to be initialised for each block in the features and computing all features is quite tedious and always fails for the cookbook example. So we decided it's better to save all the output figures instead.

pfebrer:
Ok, I see, I think the end of the notebook needs a bit more explaining.

There is a section called "Evaluating the trained model", which would suggest that you evaluate the model that has been trained, but instead you load a model called best-model that is pretrained, without any explanation. I think it would be good to add some lines there on what you are doing.

And later, even if you don't generate the plots, it will be good to have the code that generates the plots there, just as you have the code to initialize and load the model.

DivyaSuman14:
Yeah I agree, we can explain a bit more why we loaded the saved model and also add the script to generate plots.

DavideTisi (approved):

PR #146, opened 2025-06-11
Fix gaas-map (and broken website!)

ceriottm:


ceriottm:
I'm merging this quickly because the website is broken.

pfebrer:
Oops sorry my bad I guess. In which way was it broken? I can't understand it by looking at your change. Or is the build-and-publish fix unrelated to the broken website?

pfebrer:
Aah ok it was publishing even though the generate examples failed because the `conclusion` attribute didn't exist I imagine. But now the PET-MAD example seems to be failing 😅

PR #149, opened 2025-06-12
Update deps

ceriottm:
A batch of environment.yml updates reflecting the fact that a lot of the development code has become stable-ish

### DIFF
--- a/examples/lode-linear/environment.yml
+++ b/examples/lode-linear/environment.yml
@@ -3,7 +3,6 @@ channels:
 dependencies:
   - python=3.11
   - pip
-  - rust >=1.65
   - pip:
     - ase==3.22.1
     - equisolve @ git+https://github.com/lab-cosmo/equisolve.git@c858bedef4b2799eb445e4c92535ee387224089a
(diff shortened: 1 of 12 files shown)


ceriottm:
This should also accelerate tests a bit because code doesn't need to be compiled

examples/lode-linear/environment.yml
```diff
@@ -3,7 +3,6 @@ channels:
 dependencies:
   - python=3.11
   - pip
-  - rust >=1.65
   - pip:
     - ase==3.22.1
     - equisolve @ git+https://github.com/lab-cosmo/equisolve.git@c858bedef4b2799eb445e4c92535ee387224089a
```
Luthaf, line 8: we should get rid of this dependency, nobody will be able to use it …

ceriottm: I very much agree, but I think this, as well as the older Hamiltonian examples, should just be "retired" as soon as we have good examples based on the more modern technology. for the moment I'd leave them be, certainly for the purpose of this PR that is just a superficial cleaning of stuff that relies on commit pinning for reasons that are no longer relevant

PR #153, opened 2025-06-23
Use hybrid/scaled for mts in lammps

Luthaf:
Follow up to #145

### DIFF
--- a/examples/pet-mad-nc/data/lammps-respa.in
+++ b/examples/pet-mad-nc/data/lammps-respa.in
@@ -9,13 +9,18 @@ mass 2 35.453
 mass 3 1.00784
 mass 4 14.0067
 
-pair_style hybrid/overlay metatomic_1 pet-mad-latest.pt device cpu non_conservative on scale 1.0 metatomic_2 pet-mad-latest.pt device cpu non_conservative on scale -1.0 metatomic_3 pet-mad-latest.pt device cpu non_conservative off scale 1.0
+pair_style hybrid/scaled &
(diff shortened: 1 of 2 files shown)


ceriottm:
so this is just to simplify the interface on the fix_metatomic front?

Luthaf:
Yes, the `scale` parameter is only here because `hybrid/scaled` is not yet supported with kokkos (https://github.com/lammps/lammps/issues/3492), but hopefully it will be at some point and then we can remove the `scale` parameter in `pair_style metatomic`.

ceriottm:
so shall we keep this open until https://github.com/lammps/lammps/issues/3492 is merged and ported into lammps-metatomic? no strong feelings, do whatever you think makes most sense. FWIW, the current version runs smoothly on my end.

Luthaf:
I'd rather merge this now so we push people toward the correct use.

If someone wants to also use kokkos + respa, they can use the `scale` parameter!

Luthaf:
BTW, this is one of the slowest example right now, and will likely fail the 10min timeout in #155. Should I reduce the number of steps we use when running the different MD?

ceriottm:
ok, go ahead and merge. it's already doing very few steps, reducing it further will make it really meaningless.  can we carve out an exception or increase the limit to 15'?

Luthaf:
>  can we carve out an exception or increase the limit to 15'?

I would prefer to have the same time limit for everything, and 10" is already a bit long IMO. But I can merge this for now, and we can discuss the timeout later!

PR #154, opened 2025-06-23
Fail the build if a script create files that are not in gitignore

Luthaf:


pfebrer (commented):
This is cool!!

But I think it would be better to do the check in `noxfile.py`, so that the example would fail also locally.

pfebrer:
By the way, I see that you silently introduced `fail-fast: false` on the examples matrix. I think it is a good idea, but perhaps something that needs the approval of @ceriottm :sweat_smile:

ceriottm:
> By the way, I see that you silently introduced `fail-fast: false` on the examples matrix. I think it is a good idea, but perhaps something that needs the approval of @ceriottm :sweat_smile:  

What does fail-fast do?

Luthaf:
Not silently, there is a commit just for it ^_^

Basically, with `fail-fast: true` if any example fail CI will stop trying to build the other ones. So I changed it to `fail-fast: false` and it continues building other examples when one fails

Luthaf:
> But I think it would be better to do the check in noxfile.py, so that the example would fail also locally.

Good point!

ceriottm:
> Not silently, there is a commit just for it ^_^
> 
> Basically, with `fail-fast: true` if any example fail CI will stop trying to build the other ones. So I changed it to `fail-fast: false` and it continues building other examples when one fails

OK very much agree with this - it's painful to debug one example at a time, and as long as we don't pay for CI time....

Luthaf:
> But I think it would be better to do the check in noxfile.py, so that the example would fail also locally.

Actually, this could be more annoying than helpful: when working on the example, you don't want to be modifying the gitignore every time, and you also don't want to have the doc build fail because there is some extra file in the repo …

pfebrer:
Hmm ok I see, but on the other hand doing it only in CI can be a bit useless because the developer might already have pushed the unwanted files.

How does the typical workflow for developing the examples work? I would imagine you develop your example and only when you have it in an already close to finished state you run it with `nox`. Then it shouldn't be too annoying.

If not a failure, perhaps locally we could have a very big and visible warning, so that things build succesfully but the developer still adds the files to the gitignore.

As annoying as the local failure can be, I think we can not underestimate how annoying it is that things are green locally but fail on the CI. Perhaps whether the example fails or just warns could be controlled with a `nox` option:

```
nox -e pet-mad-nc --untracked-files-ok
```

(or the opposite, of course, I don't know what the best default would be)

pfebrer:
Another option would be to keep a registry of all this locally and make it impossible for the unwanted files to be commited, but this would require something like `pre-commit` and also it would be more involved.

Luthaf:
> Hmm ok I see, but on the other hand doing it only in CI can be a bit useless because the developer might already have pushed the unwanted files.

It is annoying that stuff would be green locally and fail on CI, so we should solve that. But this PR is not about having additional files committed (this should be checked as normal by review). This PR is about making sure all generated files are in gitignore, otherwise someone else running the example will end up with a bunch of extra files that are more likely to be added with `git add --all` later.

> If not a failure, perhaps locally we could have a very big and visible warning, so that things build succesfully but the developer still adds the files to the gitignore.

I did that for now, adding the warning at the very end of the output. I tried to make the warning hard to ignore, if it's not enough we will catch it in CI.

> How does the typical workflow for developing the examples work? I would imagine you develop your example and only when you have it in an already close to finished state you run it with nox.

I tend to run nox relatively early and interactively add to the example. Most of this happen with new files for an example not committed yet, so a hard failure would be quite annoying as well, especially for people less comfortable with git.

pfebrer (approved):

PR #155, opened 2025-06-23
Make CI faster

Luthaf:


.github/workflows/docs.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         run: python -m pip install nox docutils
 
       - name: build example
+        timeout-minutes: 10
```
ceriottm: I'd make this 16' so it doesn't fail current examples ^_^

src/get_examples.py
```diff
@@ -67,7 +68,7 @@ def create_ci_jobs_json(modified_files: str):
         # by the Github Workflow as having no examples to run.
         return ""
 
-    return json.dumps({"example-name": examples})
+    return json.dumps({"example-name": list(sorted(examples))})
```
pfebrer, line 71: Is this just so that the list of CI runs looks nicer? Or there was some deeper problem?

pfebrer:
By the way will the pytorch-cpu thing work outside of CI? I mean when users download the environnent.yaml . Also do we want the users to run the example with pytorch-cpu?

src/get_examples.py
Luthaf: Just to make the list nicer, and match the local ordering in my editor

Luthaf:
> By the way will the pytorch-cpu thing work outside of CI? I mean when users download the environnent.yaml . Also do we want the users to run the example with pytorch-cpu?

Yes, this is a bit of a downside. I'll see if we can do the same with an environment variable instead like we do for pip.

.github/workflows/docs.yml
Luthaf: I did 12min, pet-mad-nc takes 10min currently, so this should be enough of a buffer for run to run variations

Luthaf:
For pytorch-cpu: I now change the dependencies inside the noxfile, which also allows to remove the pin for cpu for lammps-metatomic.

.github/workflows/docs.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
         run: |
           conda config --set notify_outdated_conda false
-          nox -e ${{ matrix.example-name }}
+          nox -e ${{ matrix.example-name }} -- --no-website
```
pfebrer, line 65: Why is this?

Luthaf: When running `nox -e some-example` locally, we want to both generate the rst files from the python scripts, and to turn this rst into HTML for local visualization. On CI the second step is a separate CI job, so we can skip the rst => HTML transformation in this step and save some time.

pfebrer: Ok, in principle this should also work with PRs, where for untouched exemples we download the artifacts from the latest build, no?

Luthaf: Yes! The website itself is built below, after downloading all artifacts

.github/workflows/docs.yml
```diff
@@ -100,7 +105,7 @@ jobs:
           merge-multiple: true
 
       - name: build documentation
-        run: nox -e build_docs
+        run: nox -e build_website
```
Luthaf, line 117: @pfebrer here

ceriottm (approved):
LGTM (and it IS faster)

ceriottm:
doesn't build locally

PR #88, opened 2024-10-11
Running metadynamics with PLUMED and metatensor-defined collective variables

Luthaf:
<details>
Very work in progress, this will not even run on CI!

To build the example locally: 

```bash
git clone https://github.com/lab-cosmo/plumed2 --branch conda-metatensor plumed-conda-metatensor
cd plumed-conda-metatensor

# install the dependencies provided by pip
pip install cython "metatensor-torch>=0.5.5"

# SETUP FLAGS following https://www.plumed.org/doc-master/user-doc/html/_m_e_t_a_t_e_n_s_o_r_m_o_d.html, 
# the code below should work on linux, when using pre-built from pip wheels

# export the location to all the libraries:
TORCH_CMAKE_PREFIX=$(python -c "import torch; print(torch.utils.cmake_prefix_path)")
TORCH_PREFIX=$(cd "$TORCH_CMAKE_PREFIX/../.." && pwd)
 
METATENSOR_CMAKE_PREFIX=$(python -c "import metatensor; print(metatensor.utils.cmake_prefix_path)")
METATENSOR_PREFIX=$(cd "$METATENSOR_CMAKE_PREFIX/../.." && pwd)
 
METATENSOR_TORCH_CMAKE_PREFIX=$(python -c "import metatensor.torch; print(metatensor.torch.utils.cmake_prefix_path)")
METATENSOR_TORCH_PREFIX=$(cd "$METATENSOR_TORCH_CMAKE_PREFIX/../.." && pwd)
TORCH_CPPFLAGS="-D_GLIBCXX_USE_CXX11_ABI=0"

TORCH_INCLUDES="-I$TORCH_PREFIX/include -I$TORCH_PREFIX/include/torch/csrc/api/include"
CPPFLAGS="$TORCH_INCLUDES $TORCH_CPPFLAGS -I$METATENSOR_PREFIX/include -I$METATENSOR_TORCH_PREFIX/include $CPPFLAGS"
 
LDFLAGS="-L$TORCH_PREFIX/lib -L$METATENSOR_PREFIX/lib -L$METATENSOR_TORCH_PREFIX/lib $LDFLAGS"
LDFLAGS="$LDFLAGS -Wl,-rpath,$TORCH_PREFIX/lib"
LDFLAGS="$LDFLAGS -Wl,-rpath,$METATENSOR_PREFIX/lib -Wl,-rpath,$METATENSOR_TORCH_PREFIX/lib"
[...]

PR #157, opened 2025-07-01
Follow up for metatomic-plumed example

Luthaf:
This is a follow up to #88 with a bunch of small changes and improvements.


examples/metatomic-plumed/metatomic-plumed.py
```diff
@@ -1,55 +1,49 @@
...
-this exanmple as a template, you should set more appropriate parameters.
+2.  Implement these custom CV using ``featomic``, ``metatensor``, and ``metatomic`` to
+    create a portable ``metatomic`` model.
+3.  Run metadynamics trajectories with LAMMPS, and visualize the system as it explores
+    different configurations.
+4.  Show an example of integration with `i-PI <https://ipi-code.org/>`_, that uses
+    multiple time stepping to reduce the cost of computing complicated CVs.
+
+As usual for these examples, the simulation is run on a small system and for a short
+time, so that results will be fast but inaccurate. If you want to use this exanmple as a
```
HaoZeke:
```suggestion
time, so that results will be fast but inaccurate. If you want to use this example as a
```

HaoZeke (commented):
One spelling nit and whatever's keeping CI sad, otherwise LGTM.

ceriottm:
this doesn't run because it tries to output to out/ and the folder doesn't exist. give me 3'

ceriottm:
LOL using the neighbor list seems much slower than using torch.cdist -- even on a CPU - FTR full CI run was taking 4' in the previous runs

Luthaf:
> LOL using the neighbor list seems much slower than using torch.cdist -- even on a CPU - FTR full CI run was taking 4' in the previous runs

Yeah, this is bad 😅. I guess `index_add_` does more work than just summing over the columns. I hope this is just a side effect of having relatively few atoms in the system.

ceriottm:
> > LOL using the neighbor list seems much slower than using torch.cdist -- even on a CPU - FTR full CI run was taking 4' in the previous runs
> 
> Yeah, this is bad 😅. I guess `index_add_` does more work than just summing over the columns. I hope this is just a side effect of having relatively few atoms in the system.

I guess it's a lot of overhead to make a neighbor list for 38 atoms, and probably it has to compute all distances because the cluster is so small and compact that the linked cells will not work very well. I think it's still worth to have it slower to show an example of this, it'll be useful in many other cases. another thing to speed up though ^_^

ceriottm (approved):
LGTM

PR #158, opened 2025-07-02
Update metatomic-plumed example to no longer use LJ units

Luthaf:


ceriottm (approved):
Did a few minor tweaks (most notably the sigma of the SOAP features had not been adapted to Ar length scale). All looks good now.

PR #161, opened 2025-07-10
Update non-conservative PET-MAD recipe to mention kokkos

frostedoyster:
As in the title


ceriottm:
(1) I'd have just one example of running LAMMPS with kokkos (maybe the MTS one) 
(2) for subprocess.run you need to  split the string
(3) does one have to ask for gpu in the lammps input too?

Luthaf:
>  (3) does one have to ask for gpu in the lammps input too?

Do you mean with `device cuda`? In kokkos mode, pair_style metatomic will automatically use the same device that is used by kokkos.

ceriottm:
> > (3) does one have to ask for gpu in the lammps input too?
> 
> Do you mean with `device cuda`? In kokkos mode, pair_style metatomic will automatically use the same device that is used by kokkos.

I mean that on the command line of lammps in the recipe there is no reference to the device so I don't see how kokkos will realize it should use GPUs. at the very least, it'd be useful to explain as at present the text suggests that one should change device mode for pair_metatensor, and you seem to imply the opposite

Luthaf:
Right, one needs to ask for GPU at the command line with `lmp -in input-file -k on g 1` (for 1 GPU) or `lmp -in input-file -k on g 4` (for more GPUs). See https://docs.lammps.org/Speed_kokkos.html#running-on-gpus

ceriottm:
so @frostedoyster I am not sure of what's going on here. I don't think we can _run_ the kokkos-enabled version in the CI because we only build GPU-kokkos if I understand correctly and the version we install in environment.yml doesn't have kokkos. also I thought the whole thing holding us back was about hybrid/overlay vs hybrid/scaled, but there's inconsistency between what's written in the recipe and what's in lammps-respa.in. 
could you look after those things, and just have a final section in the recipe that explains how to install lammps-kokkos (or point to the metatomic page) and how to run with it?

PR #165, opened 2025-07-24
Add token to github API call in CI

pfebrer:
Every once in a while the CI fails because we are sending an authenticated request to the github API, which has rate limits.

In this PR I use our GH_READ_TOKEN for authenticating so that this doesn't happen anymore

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,6 +15,7 @@ jobs:
     outputs:
       examplesmatrix: ${{ steps.gatherInfo.outputs.examplesjson }}
       latest_docs_run: ${{ steps.gatherInfo.outputs.latest_docs_run }}
+    permissions: read-all # Permissions for the GITHUB_TOKEN
     steps:
       - uses: actions/checkout@v4
         with:
@@ -33,7 +34,7 @@ jobs:
             GET_EXAMPLES_ARGS=" --modified-files $(git diff --name-only -r HEAD^1 HEAD)"
           fi
           echo examplesjson=$(./src/get_examples.py ${GET_EXAMPLES_ARGS}) >> $GITHUB_OUTPUT
-          echo latest_docs_run=$(./src/latest_docs_run.py id) >> $GITHUB_OUTPUT
+          echo latest_docs_run=$(./src/latest_docs_run.py id --api-token ${{ secrets.GITHUB_TOKEN }}) >> $GITHUB_OUTPUT
 
   generate-example:
     needs: setup
@@ -83,6 +84,8 @@ jobs:
     # Run this job even if the generate-example job was skipped due to
     # no examples to run. I.e. only avoid if generate-example failed.
     if: ${{ always() && needs.generate-example.result != 'failure' }}
+    # Permissions for the GITHUB_TOKEN (does not affect GH_DEPLOY_TOKEN)
+    permissions: read-all
     needs: [setup, generate-example]
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


Luthaf:
How will this interact with PR from a fork? Could we use the Github Action-provided `GITHUB_TOKEN` instead, which should be always available? https://docs.github.com/en/actions/tutorials/use-github_token-in-workflows

pfebrer:
Yes you are right, sounds good! Do you know which rights does GITHUB_TOKEN have? Because if it's just "read" we can also use it for download-artifact

Luthaf:
From https://github.com/orgs/community/discussions/25358, it does seem that the token is set with read-only permission

pfebrer:
Aaah ok, so it's not ${{ secrets.GITHUB_TOKEN }} but simply $GITHUB_TOKEN, right?

Luthaf:
I think it's still `${{ secrets.GITHUB_TOKEN }}`, but it is automatically added by GH and we don't need to create it ourself

pfebrer:
Ok, got it! I think now should be fine. I found that one can set the permissions of GITHUB_TOKEN for each job step, so I set it to `read` and passed the GITHUB_TOKEN to download-artifact as well, let's see if it works :crossed_fingers:

pfebrer:
Argh annoying, one can only set the permissions for a job, not a job step. And we need write permissions to publish to gh-pages...

pfebrer:
The only solution I can think of is to have a secret token that can write, and use it for the deployement to gh-pages

Luthaf:
Yes, using a different token for the final deployment sounds good

pfebrer:
Ok, so let me know when you create it and what's the name of it and I will try. I'm not sure though if the "permissions" thing will also apply to this new token, and this will not come up until the PR is merged and the CI tries to deploy :sweat_smile:

.github/workflows/docs.yml
```diff
@@ -27,13 +27,14 @@ jobs:
       - id: gatherInfo
         # Collect the examples that need to be run, and the id of the latest
         # workflow run on the main branch.
+        permissions: read
```
Luthaf: Is this required? I think `read` is the default anyway

.github/workflows/docs.yml
```diff
@@ -99,10 +100,11 @@ jobs:
       - name: Download latest main branch built examples
         uses: actions/download-artifact@v4
         if: ${{ github.event_name == 'pull_request' }}
+        permissions: read
```
Luthaf: same

Luthaf (commented):
I'll generate a token now

Luthaf:
You can use `GH_DEPLOY_TOKEN` for this. You can also temporarily remove the `if: github.event_name == 'push'` condition on the gh -pages deployment to try it in this PR

.github/workflows/docs.yml
pfebrer: Hmm no read is not the default, because GITHUB_TOKEN until now has been used to deploy to gh-pages. In PRs that come from forks however GITHUB_TOKEN can only read, so the protection against attacks from PR submitters is already there by default, but not against github actions having some malicious code.

Luthaf: ok, fair enough! Although the only scenario where I would see someone trying to use malicious code would be a PR from a fork anyway

pfebrer: Well, the code in a Github Action can do whatever it wants really, so if you give them the token they could in principle do whatever they want with your repo :sweat_smile:

pfebrer:
Oops, it failed. I don't know if because `permissions` applies to all tokens or because the permissions for `GH_DEPLOY_TOKEN` are not set correctly

pfebrer:
Ok, apparently the permissions of GH_DEPLOY_TOKEN are not correct somehow :sweat_smile:

Luthaf:
I updated the token, it seems to be working now!

pfebrer:
All good, ready to merge!

Luthaf (approved):

pfebrer:
Thanks @sofiia-chorna and @johannes-spies for bringing up the API rate limits and the problems with the github tokens in forks, the problems should be solved now! :)

PR #166, opened 2025-07-25
add petmad category

DavideTisi:


ceriottm:
Maybe we can call it "Universal ML models" so it is less narrow? I agree that there's a growing number of PET-MAD-based examples, but still. That way, we can also include also FlashMD

PR #167, opened 2025-07-29
[ENH] Transition states with EON and PET-MAD

HaoZeke:
More authors / comments welcome. WIP.


HaoZeke:
cc @HannaTuerk @abmazitov for a first rough draft of the text

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,676 @@
...
+reaction like this, a basic NEB implementation can struggle to converge
+or may time out. We will show how EON's advanced features, such as
+**energy-weighted springs** and mixing in **single-ended dimer search steps**, can
+efficiently locate and refine the transition state along the path.
+
+Our approach will be:
+
+1. Set up the **PET-MAD metatomic calculator**.
+2. Illustrate the limitations of a standard NEB calculation in ASE.
+3. Use ASE to generate an initial IDPP reaction path.
```
HannaTuerk:
```suggestion
2. Use ASE to generate an initial IDPP reaction path.
3. Illustrate the limitations of a standard NEB calculation in ASE.
```

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,676 @@
...
+print(result.stdout)
+if result.stderr:
+    print(result.stderr)
+
+display(Image("1D_oxad.png"))
+
+
+# %%
+# Also, the PES 2D landscape profile as a function of the RMSD [2] which shows
+# the relative distace between edpoints as the optimization takes place:
```
HannaTuerk:
```suggestion
# the relative distance between the endpoints as the optimization takes place:
```

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,676 @@
...
+#
+# TODO(rg): this needs IRA within the environment which is not pip installable..
+# os.environ.pop("MPLBACKEND", None)
+# """
+# --con-file neb.con --plot-structures "crit_points" --facecolor "white"
+# --rc-mode path --ase-rotation="90x,0y,0z" --title "NEB-RMSD Surface"
+# --fontsize-base 16 --landscape-mode "surface" --landscape-path
+# --"all" --plot-type "landscape" --output-file 2D_oxad.png --show-pts True
+# --surface-type "rbf" --theme ruhi
+# """
```
HannaTuerk: doesn't the end point relaxation below use ira as well and it works?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,676 @@
...
+
+display(Image("2D_oxad.png"))
```
HannaTuerk: not properly displayed

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,676 @@
...
+# Finally we can visualize these with ``chemiscope``.
+#
+# settings = {
+#    "structure": [
+#        {"playbackDelay": 50, "unitCell": True, "bonds": True, "spaceFilling": True}
+#    ]
+# }
+# chemiscope.show([reactant, product], mode="structure", settings=settings)
+# %%
+# or with ASE.
```
HannaTuerk:
```suggestion
# Finally we can visualize these with ASE.
```

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Yup, need to rethink the design a little, the script dispatches through a different Python interpreter, so it's kinda a bug in `rgpycrumbs`

.gitignore
```diff
@@ -8,3 +8,238 @@ docs/src/_static/all-examples-data.js
...
+# PyPI configuration file
+.pypirc
+
+# Marimo
+marimo/_static/
+marimo/_lsp/
+__marimo__/
+
+# Streamlit
+.streamlit/secrets.toml
```
pfebrer: I would rather keep the most minimal `.gitignore` possible, why did you need this?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 nox
 docutils
 pyyaml
+requests
```
pfebrer, line 4: Why is this needed here?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+   method.
+5. Visualize the final converged pathway.
+6. Demonstrate endpoind relaxation with EON
+
+
+Importing Required Packages
+---------------------------
+First, we import all the necessary python packages for this task.
+By convention, all ``import``
+statements are at the top of the file.
```
pfebrer, line 42: Do we need to explicitly state this convention? :laughing:

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+# connecting the configurations (each point representing a full molecular
+# structure) by a discrete set of images allows one to evolve the path
+# under an optimization algorithm, and allows approximating the reaction to
+# three states, the reactant, product, and transition state.
+# The location of this transition state (≈ the point with the highest
+# energy along this path), which is relevant to determine the barrier
+# height of the relevant reaction, and can be found
+# by applying a transformation to the second
+# derivatives, to enable optimization algorithms to step along the
+# softest mode to reach a saddle configuration. An approximation which is free
```
pfebrer: I don't understand this sentence, I think the grammar is wrong (?)
> 
> The location of this transition state (≈ the point with the highest
> energy along this path), which is relevant to determine the barrier
> height of the relevant reaction, and can be found
> by applying a transformation to the second
> derivatives, to enable optimization algorithms to step along the
> softest mode to reach a saddle configuration.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+ax1.text(0.3, -1, "reactant")
+ax2.text(0.3, -1, "product")
+ax1.set_axis_off()
+ax2.set_axis_off()
+
+# %%
+# Endpoint minimization
+# ~~~~~~~~~~~~~~~~~~~~~
+#
+# For finding reaction pathways, the endpoints should be minimized.  We provided
```
pfebrer: "We provide" ?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+# initial configurations which are already minimized, but in order to see how to
+# relax endpoints with EON, please have a look at the end of this tutorial.
+
+
+# %%
+# Initial path generation
+# ~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The earliest NEB methods linearly interpolate between the two known
```
pfebrer: Technically, the linear interpolation is already the NEB method or just a method to get initial configurations?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+neb = NEB(images)
+neb.interpolate("idpp")
+
+# %%
+# We don't cover subtleties in setting the number of images, typically too many
+# intermediate images may cause kinks but too few will be unable to resolve the
+# tangent to any reasonable quality.
+
+
+# %%
```
pfebrer: Do we need a separate `%%` section here?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+    for filepath in output_files:
+        abs_path = os.path.abspath(filepath)
+        f.write(f"{abs_path}\n")
+
+print(f"Wrote absolute paths to '{summary_file_path}'.")
+
+# %%
+# ASE
+# ---
+#
```
pfebrer, line 236: From here now you are running the actual NEB with ASE and EON, no? I think a section header like `Running NEB` is needed.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+# In the 100 NEB steps we took, the structure did unfortunately not converge.
+# The metatomic calculator for PET-MAD v1.0.2 provides `LLPR based energy
+# uncertainities <https://atomistic-cookbook.org/examples/pet-mad-uq/pet-mad-uq.html>`_.
+# As we obtain a warning that the uncertainty of the path structure sampled is
+# very high, we stop after 100 steps.
+# The ASE algorithm with LBFGS optimizer does not
+# find good intermediate structures and does not converge
+# at all. Our test showed that the FIRE
+# optimizer works better in this context, but still takes over 500 steps
+# to converge.
```
pfebrer: If the FIRE optimizer works better, wouldn't it be more fair for ASE to use it? :sweat_smile:

pfebrer: Or at least provide the code to run with the FIRE optimizer.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+# We thus want to look at a different code, which manages to compute a
+# NEB for this simple system more efficiently.
+
+
+# %%
+# EON and Metatomic
+# -----------------
```
pfebrer: ASE is also `ASE and metatomic`, right? Maybe we should drop the metatomic name from the section?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+    "crit_points",
+    "--facecolor",
+    "floralwhite",
+    "--plot-type",
+    "profile",
+    "--ase-rotation=90x,0y,0z",
+    "--title=NEB Path Optimization",
+    "--output-file",
+    "1D_oxad.png",
+]
```
pfebrer, line 527: Maybe explain what you are doing here, at least mention that you will use `rgpycrumbs` with a one sentence description of what it is.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+
+result = subprocess.run(oneDprof_oxad, capture_output=True, text=True)
+print(result.stdout)
+if result.stderr:
+    print(result.stderr)
+
+display(Image("1D_oxad.png"))
```
pfebrer: This image is generated by the code, right? Because it has been commited to the repo.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+
+
+# %%
+# Also, the PES 2D landscape profile as a function of the RMSD [2] which shows
+# the relative distance between the endpoints as the optimization takes place:
+#
+# TODO(rg): this needs IRA within the environment which is not pip installable..
```
pfebrer: Remove TODO before merging

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,667 @@
...
+aseio.write("min_reactant/pos.con", reactant)
+write_eon_min_config(Path("min_reactant"), Path(fname).absolute())
+
+Path("min_product").mkdir(exist_ok=True)
+aseio.write("min_product/pos.con", product)
+write_eon_min_config(Path("min_product"), Path(fname).absolute())
+
+# read the minimized end points
+reactant = aseio.read("min_reactant/pos.con")
+product = aseio.read("min_product/pos.con")
```
pfebrer: Does this relaxation take long? If not, could we do it at the beginning?

pfebrer (commented):
Thanks! Here are a few comments

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Done in https://github.com/HaoZeke/rgpycrumbs/pull/18

.gitignore
HaoZeke: ah I just used a template, cleaning it up.

requirements.txt
HaoZeke: used to download the right `pet-mad` checkpoint.. xref:

```python
if not Path(fname).exists():
    response = requests.get(url)
    if response.ok:
        with open(fname, "wb") as f:
            f.write(response.content)
        print(f"Downloaded {fname} from tag {tag}.")
    else:
        print("Failed to download:", response.status_code)
else:
    print(f"{fname} from tag {tag} already present.")

if not Path(fname.replace("ckpt", "pt")).exists():
    subprocess.run(
        [
            "mtt",
            "export",
            "pet-mad-v1.0.2.ckpt",  # noqa: E501
        ]
    )
```

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: it doesn't take particularly long, but we discussed that it might be more pedagogical to not focus on the minimization in the beginning.

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: I was a bit unsure about the rules on this, my reasoning was that if the image is committed then we could eventually do some kind of image similarity to see if the recipe is somehow broken, but indeed they could just be removed from this.

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Updated to have ASE and Metatomic as the other title

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Kinda, some people prefer imports where they are used in Jupyter notebooks, e.g. `ira` isn't used until the appendix basically, but to keep clean lints it is imported at the top.

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Now broken into three

> Given two known configurations on a potential energy surface (PES),
> often one wishes to determine the path of highest probability between
> the two. Under the harmonic approximation to transition state theory,
> connecting the configurations (each point representing a full molecular
> structure) by a discrete set of images allows one to evolve the path
> under an optimization algorithm, and allows approximating the reaction to
> three states: the reactant, product, and transition state.
> 
> The location of this transition state (≈ the point with the highest energy
> along this path) determines the barrier height of the reaction. This saddle
> point can be found by transforming the second derivatives (Hessian) to step
> along the softest mode. However, an approximation which is free from
> explicitly finding this mode involves moving the highest image of a NEB path:
> the "climbing" image.
> 
> Mathematically, the saddle point has zero first derivatives and a single
> negative eigenvalue. The climbing image technique moves the highest energy
> image along the reversed NEB tangent force, avoiding the cost of full
> Hessian diagonalization used in single-ended methods [2].
>

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Depends, I mean nowadays I'd say most people wouldn't think it as part of the NEB, it's just the initial path generation method. Of course the earlier papers implicitly meant get a linear interpolation...

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: part of the reasoning is that EON uses the LBFGS as well

requirements.txt
pfebrer: But isn't this for the recipe (instead of a general dependency)?

examples/eon-pet-neb/eon-pet-neb.py
pfebrer: Ah ok, well if we want to do this at some point we could also just store the last build and compare between builds. I would say for now we can leave it out.

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: added a line about it

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Clarified that second order methods are nicer

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Especially because there's then the discussion on IRA which might break the flow.

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Rewrote this to be a bit clearer.

requirements.txt
HaoZeke: ah no, my bad.. so this is a weirdly "hidden" dependency. For instance:

```bash

❯ ag requests
examples/batch-cp2k/reference-trajectory.py
32:import requests
171:        response = requests.get(url)
55:    response = requests.get(url)

examples/hamiltonian-qm7/environment.yml
examples/gaas-map/gaas-map.py
30:import requests
51:    response = requests.get(url)

examples/hamiltonian-qm7/hamiltonian-qm7.py
83:import requests
193:    response = requests.get(url)

examples/lpr/lpr.py
21:import requests
43:    response = requests.get(url)

examples/periodic-hamiltonian/periodic-hamiltonian.py
25:import requests
282:    response = requests.get(url)

examples/periodic-hamiltonian/environment.yml
13:      - requests

examples/pet-mad/pet-mad.py
59:import requests
98:    response = requests.get(url)

examples/sample-selection/sample-selection.py
24:import requests
46:    response = requests.get(url)

examples/eon-pet-neb/eon-pet-neb.py
56:import requests
88:    response = requests.get(url)

examples/shiftml/shiftml-example.py
18:import requests
36:    response = requests.get(url)

src/latest_docs_run.py
13:import requests
32:    response = requests.get(
39:            "exclude_pull_requests": True,
117:        are rate limits for unauthenticated requests.""",

requirements.txt
4:requests
```

So it is used in a bunch of recipes, but not explicitly noted before this.

HaoZeke:
Requesting another round of review, the build will work as soon as the CDN outage over at `conda-forge` is fixed..

image

([status](https://conda-forge.org/status/))

since the new EON is already uploaded..
image

requirements.txt
pfebrer: Ok, so it is used in `src/latest_docs_run.py`, so fine to add it here. Then we should add the `requests` dependency to all those recipes, no?

pfebrer: In principle the base environment should not be used by the recipes as far as I understand

HaoZeke: > In principle the base environment should not be used by the recipes as far as I understand

Yes but this is currently not true at all, xref prior conversations about `chemiscope` leaking everywhere and #190

pfebrer (approved):

requirements.txt
pfebrer: Yes but we should have the dependencies specified just in case we want to make it right at some point haha

HaoZeke: Absolutely, but to keep this one clean I'll do it in a follow up PR ^_^

abmazitov (approved):
Fantastic job @HaoZeke and @HannaTuerk , thank you a lot!

requirements.txt
HaoZeke: xref https://github.com/lab-cosmo/atomistic-cookbook/pull/193

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -0,0 +1,790 @@
...
+# %%
+# Obtaining the Foundation Model - PET-MAD
+# ----------------------------------------
+#
+# ``PET-MAD`` is an instance of a point edge transformer model trained on
+# the diverse `MAD dataset <https://arxiv.org/abs/2506.19674>`__
+# which learns equivariance through data driven measures
+# instead of having equivariance baked in. In turn, this enables
+# the PET model to have greater design space to learn over. Integration in
+# Python and the C++ EON client occurs through the ``metatomic`` software,
```
HannaTuerk:
```suggestion
# Python and the C++ EON client occurs through the ``metatomic`` software [1],
```

HannaTuerk (approved):

HaoZeke:
OK I'm squashing this in at long last, since #194 is to be handled separately after https://github.com/metatensor/plumed-metatomic-feedstock/pull/6 and a companion to the `py-plumed` thing (not sure we need that yet).

Thanks for all the work @HannaTuerk @abmazitov @pfebrer.

PR #169, opened 2025-08-06
Electronic heat capacity with PET-MAD and PET-MAD-DOS

pfebrer:
This is an example combining PET-MAD with PET-MAD-DOS to compute the electronic heat capacity of a material.

### DIFF
--- a/examples/pet-mad-dos/pet-mad-dos.py
+++ b/examples/pet-mad-dos/pet-mad-dos.py
@@ -16,72 +16,76 @@ In the example, we will:
 """
 
 # %%
+# Import necessary libraries
+# ---------------------------
 #
-# Load PET-MAD
-# ------------
-#
+
+from collections.abc import Sequence
 
 import numpy as np
 import torch
-from pet_mad.calculator import PETMADCalculator
 
+import ase
+from ase.build import bulk
+import ase.md
+from ase.units import kB
+
+from pet_mad.calculator import PETMADCalculator, PETMADDOSCalculator
+
+import matplotlib.pyplot as plt
+
+# %%
+#
+# Load the PET-MAD models
+# -----------------------
+#
+# Here we load two models from PET-MAD:
+#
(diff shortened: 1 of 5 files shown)


examples/pet-mad-dos/pet-mad-dos.py
```diff
@@ -0,0 +1,275 @@
...
+    # and find the one that gives the correct number of electrons.
+    integrated_doses = []
+    trial_fermis = np.linspace(Ef_0 - 0.5, Ef_0 + 0.5, 100)
+    for trial_fermi in trial_fermis:
+        fd = fd_distribution(E, trial_fermi, T)
+        integrated_dos = torch.trapezoid(dos * fd, E)
+        integrated_doses.append(integrated_dos)
+
+    n_elecs_Ef = interp1d(integrated_doses, trial_fermis)
+    Ef = n_elecs_Ef(n_elec)
```
HowWeiBin, line 177:
```suggestion
    try:
        Ef = n_elecs_Ef(n_elec)
    except:
        Ef = Ef_0
        print ("Finite temperature Fermi Determination failed, defaulting to 0K Fermi level")
```

pfebrer: I think it's a bad idea to just print a warning and return something that the user has not asked for, I think the best idea is to raise an error (be it the one that is naturally raised or one with a better message) and let the user handle the situation as they wish

pfebrer: Sorry I thought this was the pet-mad repo and therefore for users to use, but anyway, I think it should just error if there is an error

pfebrer:
This is stalled right now mainly because I'm not sure that electronic heat capacity is the best tutorial to have on the PET-MAD-DOS front, I am unsure of how interesting it is

pfebrer:
There is already a pet-mad-dos recipe

PR #181, opened 2025-10-22
Update FlashMD demo to use FlashMD 0.2.1

frostedoyster:
This FlashMD version contains r2SCAN models and uses the new metatomic standard outputs for positions and momenta

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -58,7 +58,7 @@ jobs:
         run: python -m pip install -r requirements.txt
 
       - name: build example
-        timeout-minutes: 12
+        timeout-minutes: 60
         env:
           PIP_EXTRA_INDEX_URL: https://download.pytorch.org/whl/cpu
         run: |
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -58,7 +58,7 @@ jobs:
         run: python -m pip install -r requirements.txt
 
       - name: build example
-        timeout-minutes: 60
+        timeout-minutes: 12
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: python -m pip install -r requirements.txt
 
       - name: build example
-        timeout-minutes: 12
+        timeout-minutes: 60
```
pfebrer: Not sure we want to do this, since this was a recent introduction by @Luthaf or @HaoZeke . Is it really needed?

frostedoyster: No, I'm just trying to get the example to build to see where it's taking long

pfebrer: Ah ok :+1: Can you comment this line while you do the tests (if you need to do more) https://github.com/lab-cosmo/atomistic-cookbook/blob/a9b8b017086daff85cef8ca276c0eb954ad447a5/src/get_examples.py#L17C1-L17C49 ? So that it doesn't run all the examples.

pfebrer: Also the line above that one, I didn't take into account second order effects :laughing:

frostedoyster:
All examples keep running despite the fact that I changed the timeout back (and I also followed pfebrer's advice). Closing and reopening

frostedoyster:
It's still running all the examples. I will reopen it from a different branch

pfebrer:
Yess you probably didnt see my last comment, by modifying src/get_examples.py you also trigger all runs, you just needed to comment as well the line above the one that you commented 😅

PR #183, opened 2025-10-25
Remove warmup epochs from training options

frostedoyster:
Fixes the `pet-finetuning` example


pfebrer:
Nice thanks, so the effect of removing the input is the same as the previous `num_epochs_warmup: 1`?

pfebrer:
Should we update the `metatrain` dependency to only the versions with the new scheduler?

PR #184, opened 2025-10-25
Fix Github API authentication

pfebrer:
In the requests to the Github API sent directly by our code, I thought we where authenticating but we were not.

Now it is really following the instructions in : https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?utm_source=chatgpt.com&apiVersion=2022-11-28#about-authentication.

If the authentication fails we will get a `401: Bad credentials` error.

### DIFF
--- a/src/latest_docs_run.py
+++ b/src/latest_docs_run.py
@@ -25,14 +25,18 @@ def get_latest_successful_docs_run(api_token: Optional[str] = None) -> int:
 
     doc_runs_endpoint = GITHUB_ACTIONS_API + "/workflows/docs.yml/runs"
 
+    headers = {}
+    if api_token is not None:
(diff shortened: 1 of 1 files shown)


pfebrer:
The CI here will fail until the examples `pet-finetuning` and `pet-mad-uq` are fixed.

pfebrer:
I cancelled the running of the examples, and the important part for this PR (which was checking that the authentication works [here](https://github.com/lab-cosmo/atomistic-cookbook/actions/runs/18801533809/job/53649709497?pr=184#step:3:7)) is fine, we are getting the right `run_id` [here](https://github.com/lab-cosmo/atomistic-cookbook/actions/runs/18801533809/job/53649741973?pr=184#step:5:6). So this can be merged.

PR #187, opened 2025-11-02
Cleaned up a little the UQ example, so that one can see the RDF errors

ceriottm:
The RDF are also clearly scaled wrong but I can't figure out how they should be. 
Maybe we shouldn't use ASE to compute them, but this is possibly for a separate PR


johannes-spies (approved):
Looks good to me! The gaussian smoothing part of the RDF is a bit rough, but I think it's justified given the short trajectory.

PR #196, opened 2025-12-04
bug(tmp): nuke mta-plumed for a bit

HaoZeke:
The nuclear option. @pfebrer we should merge this one in today if #195 can't be resolved soon (but @Luthaf is taking a look so let's hold off until the end of the day).


HaoZeke:
obviously the intent is to revert this later, so:

```bash
git revert 9ee3c09483c17dd62673c1fea9a322526bca4bba
# make real fixes here
git commit . -m "feat(mtaplumed): fixed annoying cross version bug"
git push
# request review & profit
```

Luthaf:
ok, I think we can merge this for now, I don't understand what's wrong with the example. It looks like some kind of repeatable memory corruption, making plumed try to execute the model with `per_atom=True`, even when loading the model from Python or C++ the capabilities are properly set with `per_atom = false`.

PR #199, opened 2025-12-12
Add back metatomic-plumed

Luthaf:


Luthaf:
Well, now everything works ^_^. I think we still had some mismatched dependencies between conda and pip.

pfebrer (approved):

PR #200, opened 2025-12-18
Remove usage of `chemiscope.ase_merge_pi_frames`

Luthaf:
It required different tweaks in the two places it was used, and now the example show better how to use custom shapes for advanced visualization.


ceriottm (approved):
I trust you to check that it preserves the functionality, but I agree with the spirit of it, even if irrationally it makes me sad :-(

PR #203, opened 2025-12-29
Fix PET-MAD model save

abmazitov:


examples/pet-mad/pet-mad.py
```diff
@@ -134,23 +134,23 @@
...
+calculator = PETMADCalculator(version="1.0.2", device="cpu")
 
 # %%
 #
 # The model can also be exported in a format that can be used with
 # external MD engines. This is done by saving the model to a file,
 # which includes the model architecture and weights.
 
-calculator._model.save("pet-mad-v1.1.0.pt")
+calculator.calculator._model.save("pet-mad-v1.0.2.pt")
```
pfebrer, line 145: Isn't there a public method to do this? Guillaume said there was one

pfebrer (commented):
Thanks!

examples/pet-mad/pet-mad.py
pfebrer: As I understood, it is a function in `pet-mad`

abmazitov: There is a function indeed, but it's not public, and it actually fetches the model for a second time. I thought it's better to go with this save method

pfebrer: Hmm if this is the preferred method, should the petmad calculator have a public `model` attribute?

abmazitov: Hard to tell, the private attribute comes from the metatomic ASE calculator

pfebrer: Hmm ok, I'll leave it here so that we can discuss with Guillaume what's best. But using private things in tutorials is not nice, it gives the impression that the API is not well thought, and the tutorial is more likely to break.

If the lpr example gets fixed and this becomes the blocker I will merge as is.

cclauss:
~Instead of downgrading, why not:~
* ~#205~ (closed)

examples/pet-mad/pet-mad.py
Luthaf: IMO the best solution here is to use `pet_mad.save_pet_mad`, which is public and explicitly made for this use case. Regarding the duplicated download, doesn't pet-mad already cache downloads?

abmazitov: > IMO the best solution here is to use `pet_mad.save_pet_mad`, which is public and explicitly made for this use case. Regarding the duplicated download, doesn't pet-mad already cache downloads?

I'm afraid `pet_mad.save_pet_mad` is not a public method. We only have `pet_mad._models.save_pet_mad` available, and this is the reason why I switched to this `calculator.calculator._model.save` way. Since we use the `load_model` from metatrain inside, the checkpoint indeed should be cached I suppose. We can add the `get_pet_mad` and `save_pet_mad` to the `__init__.py` in the pet-mad repo, but that would require us making another release before merging this PR

Luthaf: `pet_mad.save_pet_mad` is documented in the readme: https://github.com/lab-cosmo/pet-mad?tab=readme-ov-file#evaluating-pet-mad-on-a-dataset, I did not check the `__init__.py`.

I'd say we should expose this function publicly ASAP, and use `pet_mad._models.save_pet_mad` in the mean time.

pfebrer: It looks like it was (mistakenly?) removed from the `__init__` here: https://github.com/lab-cosmo/pet-mad/commit/02236c1277e9e6ab4c0c920bf2279ae9158173f5

ceriottm: I'd merge this for the moment and aim to make save_pet_mad public at the next release, then update the examples. @abmazitov  can you see to this?

pfebrer: We are still waiting for the lpr example, could we make a petmad release in the meantime or would that be problematic?

pfebrer:
Since the option to use temporarily `pet_mad._models.save_pet_mad` sounds equally sketchy to me I will merge as is, but we should really update `petmad` as soon as possible, because there are two tutorials (petmad's README and this one) showing different approaches for saving, and none of them is right at the moment :sweat_smile:

pfebrer:
Thanks Arslan!

PR #209, opened 2026-01-19
Pinned the LAMMPS version to 2025.09.10.mta1 in the PET-MAD example

abmazitov:
... Until the bug introduced in `2025.09.10.mta2` is fixed


pfebrer:
Thanks, could you open an issue in the `upet` repository explaining that newer versions of lammps don't work? Just in case people use it and it fails for them, they can see that we are aware of it and the fix is to downgrade.

PR #212, opened 2026-01-28
feat(eonpet): update recipe with upet + ocineb

HaoZeke:
Cleans the recipe up to be much less scary for end-users (less boilerplate). Shifted to the `upet` stuff (needs versioning on HuggingFace).


examples/eon-pet-neb/environment.yml
```diff
@@ -3,15 +3,18 @@ channels:
...
-  - eon>=2.8.2
+  - eon>=2.9.0
   - ira
-  - libtorch>=2.9,<2.10 
-  - pip
   - python=3.13
+  - pip
   - uv
   - pip:
+    - --extra-index-url https://download.pytorch.org/whl/cpu
```
pfebrer: Why is this CPU requirement?

HaoZeke: Not really required but other wise locally it downloads a lot of extra stuff 😅

pfebrer: Ok, but this is I would say a shared problem between many recipes no? I would keep it without the extra requirement and maybe at some point find a common solution, like automatically generating an environment file for cpu and one for gpu that people can download from the website.

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -75,39 +73,34 @@
...
-    )
+repo_id = "lab-cosmo/upet"
+tag = "v1.1.0"
+url_path = f"models/pet-mad-s-{tag}.ckpt"
+fname = Path(url_path.replace(".ckpt", ".pt"))
+url = f"https://huggingface.co/{repo_id}/resolve/main/{url_path}"
+fname.parent.mkdir(parents=True, exist_ok=True)
+subprocess.run(
+    [
+        "uvx",
```
pfebrer: What is uvx and why do we need it here?

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -496,38 +387,81 @@ def run_command_or_exit(
...
+                "all",
+                "--show-pts",
+                "--surface-type",
+                "rbf",
+            ]
+        )
+    else:
+        raise ValueError(f"Unknown plot mode: {mode}")
+
+    return base_cmd
```
pfebrer: I think eventually it would be better for the tutorial if you provided a python API for these `rgpycrumbs` utils. But not something to change now.

pfebrer (changes requested):
Good, I agree that it is better, specially using dictionaries for settings instead of just multiline strings 👍 The only thing that I want to understand before merging is if the uvx thing is needed and if we can remove the cpu requirement from the environment

examples/eon-pet-neb/eon-pet-neb.py
```diff
@@ -496,38 +387,81 @@ def run_command_or_exit(
...
-    "--output-file",
-    "1D_oxad.png",
-]
-
 # Run the 1D plotting command using the helper
-run_command_or_exit(oneDprof_oxad, capture=False, timeout=60)
+cmd_1d = get_neb_plot_cmd(
+    "profile", title="NEB Path Optimization", output_file="1D_oxad.png"
+)
+run_command_or_exit(cmd_1d, capture=False, timeout=60)
```
pfebrer: Ah also maybe getting the command and running it could be put into the same function, `run_neb_plot`?

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Sorry my bad, for `mtt` no these are not needed, removing

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Tracked upstream https://github.com/HaoZeke/rgpycrumbs/issues/34

examples/eon-pet-neb/environment.yml
HaoZeke: Tracked for later in https://github.com/lab-cosmo/atomistic-cookbook/issues/213

examples/eon-pet-neb/eon-pet-neb.py
HaoZeke: Might be more overhead for users to add another function instead of the two lines? Also don't like side-effects that much :D

(but actually why not, on it)

pfebrer (approved):

pfebrer:
Thank you!

PR #215, opened 2026-02-02
Update shiftml dependency to use GitHub repository main

bananenpampe:


bananenpampe:
we are updating shiftml to avoid zenodo timouts

pfebrer:
Can't you do a release of shiftml? I think it would be better :sweat_smile:

bananenpampe:
I want to do a release when there are new models

pfebrer:
Why?

(I think tutorials should only use github versions if there is a very good reason, and there hardly ever is) :sweat_smile:

ceriottm:
I tend to agree this is fixing a (small)  bug, and there's no reason not to do a point release.

bananenpampe:
point release done and it should contain now the newest shiftml release

pfebrer:
Niceee thank you!

PR #224, opened 2026-02-19
Instanton rates recipes

litman90:
This PR creates a recipe to compute tunneling rates using ring polymer instanton rate theory.

PR #227, opened 2026-02-20
Instanton rates recipes

litman90:
this replaces #224  so it allows maintainers to edi it


litman90:
@ceriottm @DavideTisi  it should be ready now

ceriottm (approved):
Thanks @litman90!  Longer-term, it could be maybe made a bit more concise. if we use at the i-pi workshop we could collect feedback from participants. For the moment it looks good, will merge as soon as the CI is happy

PR #229, opened 2026-02-23
feat(eon-pet-neb): update for metatrain, upet, eOn, etc.

HaoZeke:
Bump version pins to `>=1.1.0` to pick up the orthogonal reaction-valley projection (progress vs perpendicular deviation) and gradient-enhanced IMQ surface. Updates plot descriptions.

The visual is also enhanced by having uncertainty intervals shown.


HaoZeke:
image


Because this makes more sense.

examples/eon-pet-neb/environment.yml
```diff
@@ -8,11 +8,14 @@ dependencies:
   - python=3.13
   - pip
   - uv
+  - jax
```
pfebrer, line 11: Why? 😅

HaoZeke: Sadly to get the even nicer surface interpolation 😅 (it's an optional dependency of `rgpycrumbs`)

pfebrer: Is it very bad without it? 

Also could it be the reason why you need to add `filelock` and get `metatrain` from a git commit?

HaoZeke: > Is it very bad without it?

Quite, since without it I don't have gradients and so the basins are wrong, and also no good uncertainity estimates (the lines). Have a reviewer who was very unhappy with the older version 😅 

> Also could it be the reason why you need to add `filelock` and get `metatrain` from a git commit?

No that's just because I like the `mtt export lab-cosmo/upet models/pet-mad***` style of writing which  I introduced in https://github.com/metatensor/metatrain/pull/1025 😄

HaoZeke: image


So this was the "before" thing (currently on the Cookbook) which was a thin plate spline, note that neither endpoint is in a state, and also there's no uncertainty contour (also it's RMSD-RMSD but that's just a coordinate shift).

pfebrer: And `filelock`? Do you understand why you need it now?

For the `mtt export` command I would vote to change it only once it is in a metatrain release :sweat_smile: It will probably be soon don't worry haha

pfebrer (approved):

HaoZeke:
Thanks @pfebrer

PR #234, opened 2026-03-06
Add MD to the fine-tuned ethanol model, to showcase stability even with a minimal amount of structures.

ceriottm:


pfebrer (approved):
Michele running MD with ASE, interesting :thinking:

PR #236, opened 2026-03-16
fix(eon-pet-neb): surface sphinx-gallery errors in CI, bump dep pins

HaoZeke:
sphinx-gallery captures script stdout/stderr internally during recipe execution. When a recipe fails, the traceback is swallowed and CI shows only "failed with exit code 2" with no useful diagnostics. Uses a hacky workaround to keep CI logs useful.

Also bump rgpycrumbs>=1.4.1 and chemparseplot>=1.4.2 to ditch the mistaken  plt_neb migration to chemparseplot.

### DIFF
--- a/src/generate-gallery.py
+++ b/src/generate-gallery.py
@@ -1,11 +1,28 @@
 import os
 import shutil
 import sys
+import traceback
 
 import chemiscope  # noqa: F401
 import sphinx_gallery.gen_gallery
 import sphinx_gallery.gen_rst
 from chemiscope.sphinx import ChemiscopeScraper
+
+
+# Monkey-patch _LoggingTee.write to echo captured output to sys.__stderr__.
+# sphinx-gallery's _LoggingTee replaces both sys.stdout and sys.stderr during
+# recipe execution, so all print output vanishes from CI. sys.__stderr__ is
+# the original fd preserved by Python at startup, immune to replacement.
+_orig_tee_write = sphinx_gallery.gen_rst._LoggingTee.write
+
+
+def _tee_write_with_ci(self, data):
+    _orig_tee_write(self, data)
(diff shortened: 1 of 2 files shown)


src/generate-gallery.py
```diff
@@ -87,4 +104,9 @@ def add_css_file(self, path):
     app = PseudoSphinxApp(example=sys.argv[1])
     sphinx_gallery.gen_gallery.fill_gallery_conf_defaults(app, app.config)
     sphinx_gallery.gen_gallery.update_gallery_conf_builder_inited(app)
-    sphinx_gallery.gen_gallery.generate_gallery_rst(app)
+
+    try:
+        sphinx_gallery.gen_gallery.generate_gallery_rst(app)
+    except Exception:
+        traceback.print_exc(file=sys.__stderr__)
+        sys.exit(1)
```
pfebrer, line 112: why not `raise`?

pfebrer (commented):
Ok it's hacky but it is well explained, good for me 👍 Just one comment

src/generate-gallery.py
HaoZeke: that gets eaten away by CI 😅

pfebrer: Even if you `raise` your own exception? I.e. not just reraise the sphix gallery one.

HaoZeke: Should we raise a new exception? I mean this isn't a library to have defined new ones I'd think?

pfebrer: No I was thinking just something like `raise RuntimeError()`

pfebrer: But it's fine

pfebrer:
Thanks!

PR #238, opened 2026-03-16
Add kaldo thermal conductivity recipe

gbarbalinardo:
## Summary

- New recipe computing lattice thermal conductivity of Si (diamond) using kaldo and PET-MAD (UPET)
- Includes structure relaxation, force constants, phonon dispersion, DOS, and thermal conductivity via BTE direct inversion
- Runs in ~8 min with pet-mad-xs and 3×3×3 supercells on CPU

dispersion_dos
cv
velocity_vs_q
phasespace
lifetime
kappa_cumulative_freq


ceriottm:
Love it! Is this ready for review?

pfebrer:
Hi @gbarbalinardo , thank you very much for the contribution! You can of course keep the temporary dir if you want but if you did it because of the CI errors, you can also just add a `.gitignore` file so that the created files get ignored ([example](https://github.com/lab-cosmo/atomistic-cookbook/blob/main/examples/flashmd/.gitignore)). We just do this check so that the created files don't get commited by mistake in the future

gbarbalinardo:
Thanks for the feedback @pfebrer and @ceriottm. I implemented the changes.

ceriottm:
OK, first round of comments.
1. Rename the example to thermal-conductivity-bte; as long as we don't have examples with another code, I don't see why you need to specify it's usink KALDO
2. IDK if you plan on providing other examples based on KALDO (maybe with more advanced stuff...) - in case, I'd be happy with having a software page dedicated to it. see how this is done for EON
3. Add references to other recipes, e.g. to optimize the structure https://atomistic-cookbook.org/examples/pet-relaxation/pet-relaxation.html or compute phonons https://atomistic-cookbook.org/examples/pet-phonons/pet-phonons.html
4. IMO it's nice to make these a bit more pedagogic - see if you want to break it down into subsections for the various steps, add a few equations explaining what is being computed, give references to the theory papers so people can understand better what they are doing
5. Make sure that the final plots are shown in the notebook. If those are saved as images, you can then load them so that they're visualized. Not only they'll look nicer, it'll also automatically give you a thumbnail in the recipe index.

gbarbalinardo:
Hi @ceriottm, thanks for the thorough review. I have addressed your comments:

1. Renamed the example to `thermal-conductivity-bte` including directory and files.  
2. Added a kALDo software page `docs/src/software/kaldo.sec`, following the eOn pattern. We do plan to add more kALDo examples in the future.  
3. Added cross references to the geometry relaxation and phonon dispersion recipes in both the introduction and the body of the script.  
4. Restructured the script to be more pedagogical, with clear subsections such as Setup, Structure relaxation, Force constants, Harmonic phonon properties, Anharmonic properties and thermal conductivity, and Convergence. 
5. Plots are now shown inline. I replaced `plot_crystal` with individual calls such as `plot_dispersion`, `plot_dos`, and `plot_vs_frequency` for heat capacity, bandwidth, and mean free path, plus a custom cumulative `kappa` vs frequency plot. Sphinx Gallery now captures all six figures natively, with the first used as the thumbnail.

Thanks!

ceriottm:
Thanks @gbarbalinardo. I'll do the final tweaks myself and merge, unless there is something that requires substantial effort. FYI, I had to pin numpy<2.4 because np.trapz has finally been removed, so you should update kaldo to use trapezoid

gbarbalinardo:
Thanks @ceriottm!

PR #249, opened 2026-04-24
Fix installation instructions

ceriottm:
Installation instructions were using the wrong syntax for conda activate.
Fixed and cleaned. Bit annoying to have two INSTALLING.rst  that differ subtly, but it does make sense to have different instructions for when the user has already fetched the recipe archive.


pfebrer:
Thanks!

PR #250, opened 2026-04-24
Clean up the pet-mad-dos recipe

ceriottm:
There are a few things that need to be improved in the PET-MAD-DOS recipe. 
Some also require improvements to upet.dos
SMALL TWEAKS:
- [ ]  Have a utility function in upet that pads a given DOS profile
- [ ]  Decrease learning rate for fine-tuning
- [x]  Chemiscope viewer 
- [ ]  use the same energy grid for plotting true and predicted DOS

LARGER THINGS (requiring possibly upet coding):
- [ ]  All the helper functions should be part of the upet.dos package. Doesn't make sense (and it's not sustainable) to have a bunch of opaque magic functions distributed with the recipe. E.g: 
- [ ] computing the energy grid and/or interpolating on a different grid; 
- [ ] computing the alignment; 
- [ ] computing the structure DOS and mask
- [ ]  The recipe starts using only numpy and then switches to torch in get_dynamic_shift_agnostic_mse, which is a bit surprising, and some users might not be used to torch (especially since the calculator exposes an interface based on ASE/numpy)

### DIFF
--- a/examples/pet-mad-dos/pet-mad-dos.py
+++ b/examples/pet-mad-dos/pet-mad-dos.py
@@ -31,6 +31,7 @@ import torch
 import urllib.request
 import os
 import subprocess
+import chemiscope
 
 # %%
 # Using PET-MAD-DOS out of the box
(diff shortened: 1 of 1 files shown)


pfebrer:
The work is done in https://github.com/lab-cosmo/atomistic-cookbook/pull/259

PR #251, opened 2026-04-24
Demonstration of the hourglass principle used by metatomic

Luthaf:


EricBoittier:
The `generate-example (metatomic-hourglass)` CI failure here (`build example` step, exit code 1) is an upstream conda channel issue, not caused by this branch.

`gromacs-metatomic 2026.0.mta3` (the only version currently in the `metatensor` channel) requires `libtorch >=2.13.0,<2.14.0a0`, while `lammps-metatomic==2026.07.04.mta1` requires `libtorch >=2.12.1,<2.13.0a0`. No currently available pair of builds shares a libtorch version, so `environment.yml` is unsatisfiable regardless of how the pins here are adjusted.

Root cause: `gromacs-metatomic-feedstock` moved to libtorch 2.13 on 2026-07-30, three days after `lammps-metatomic-feedstock`'s last bump only reached libtorch 2.12.

Filed upstream:
- metatensor/lammps-metatomic-feedstock#18 (full writeup + suggested fix)
- metatensor/gromacs-metatomic-feedstock#7 (cross-link)

This PR should be unblockable once `lammps-metatomic-feedstock` rebuilds against libtorch 2.13 to match `gromacs-metatomic-feedstock` — no changes needed here in `environment.yml`.

pfebrer:
@GardevoirX now that you have it fresh, could you modify the DPA3 model loading here so that it uses what you pushed to metatrain?

examples/metatomic-hourglass/metatomic-hourglass.py
```diff
@@ -0,0 +1,742 @@
...
+# ``metatrain``'s MACE architecture, as described in `the metatrain
+# documentation
+# <https://docs.metatensor.org/metatrain/latest/architectures/generated/mace.html>`_.
+# A "training" run with zero epochs re-packages the foundation model, and the
+# usual export then produces the same kind of metatomic model as before. The
+# dummy dataset is never used for actual training, and only needs to declare
+# the right target names.
+
+
+def get_mace() -> AtomisticModel:
```
pfebrer, line 97: @GardevoirX you can take this as an example, the DPA3 one should be very similar now.

GardevoirX: but do we need a new release of mtt first?

pfebrer: You can set the dependency to `metatrain@main` for now. At this pace, I'm sure we will release before this recipe goes in :sweat_smile:

GardevoirX:
I tried to override the metatrain of the nox environment with the main branch's, and the example worked. The CI should be happy again as soon as we have a new release of metatrain

pfebrer:
Hey, I don't think we should change the dpa3 part in the example. We should use it as Qianjun implemented it and set the metatrain dependency to `metatrain@main`. Then when we are ready to merge this one we can make a metatrain release.

pfebrer:
Oh no, `upet`'s pin of metatrain :sweat_smile:

examples/metatomic-hourglass/metatomic-hourglass.py
```diff
@@ -0,0 +1,752 @@
...
+
+engines = {
+    "ASE": run_ase,
+    "LAMMPS": run_lammps,
+    "GROMACS": run_gromacs,
+    "i-PI": run_ipi,
+    "TorchSim": run_torchsim,
+}
+
+unsupported = {("LAMMPS", "DPA3"), ("GROMACS", "DPA3")}
```
pfebrer: Why are lammps and gromacs unsupported for DPA3?

GardevoirX: no idea, in principle should work, I will have a try

GardevoirX:
it just occurred to me, would it be better if we add DFTD3 wrapper here? demonstrating that atomistic models can be post-modified through wrappers
https://docs.metatensor.org/metatomic/latest/torch/reference/wrappers.html#metatomic.torch.dftd3.DFTD3

pfebrer:
Hmm it adds a layer of complexity but I think it could indeed be nice, I'll let Guillaume decide :)

ceriottm:
> it just occurred to me, would it be better if we add DFTD3 wrapper here? demonstrating that atomistic models can be post-modified through wrappers https://docs.metatensor.org/metatomic/latest/torch/reference/wrappers.html#metatomic.torch.dftd3.DFTD3

Maybe we do a dedicated recipe? It is already a huge recipe and I fear it'd be lost in here

PR #252, opened 2026-05-04
fix(ml-mm): pin libopenblas to satisfy gmx_mpi dynamic loader

HaoZeke:
`libgromacs_mpi.so` links directly against `libopenblas.so.0` (NEEDED entry visible via objdump on the `gromacs-metatomic 2026.0.mta2` build), but the feedstock does not declare `libopenblas` as a runtime dependency. On hosts that happen to have a system-wide `/lib/x86_64-linux-gnu/libopenblas.so.0` the loader silently falls back to the system copy. GitHub Actions `ubuntu-latest` images seem to no longer ship that, so the example crashes with:

    gmx_mpi: error while loading shared libraries:
    libopenblas.so.0: cannot open shared object file: No such file or directory

Listing libopenblas in environment.yml puts the library in the conda prefix and the gromacs RPATH ($ORIGIN/../lib) resolves it correctly... in principle 😅 

=> follow up needed because `torch` otherwise wants the MKL symbol -_-

This should need a `run_exports` update to `gromacs-metatomic-feedstock/recipe/meta.yaml` if it works too.

### DIFF
--- a/examples/ml-mm/environment.yml
+++ b/examples/ml-mm/environment.yml
@@ -4,6 +4,9 @@ channels:
 dependencies:
   - chemiscope
   - metatensor::gromacs-metatomic=*=mpi_*
+  # gromacs-metatomic links libgromacs against libopenblas.so.0 but the recipe
+  # does not list it as a runtime dependency; pin it here so gmx_mpi can load.
+  - libopenblas
   - libtorch =*=cpu*
   - pytorch-cpu
   - matplotlib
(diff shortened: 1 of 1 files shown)


pfebrer (approved):

Luthaf:
> follow up needed because torch otherwise wants the MKL symbol -_-

If everything is built with conda-forge, we should be able to only pull one of these for both torch and the gromacs build

PR #253, opened 2026-05-04
fix(upet): force nvalchemi-toolkit-ops>=0.3.0 to dodge wp.vec regression

HaoZeke:
All four upet-based examples (`pet-phonons`, `pet-mad-dos`, `pet-relaxation`, `thermal-conductivity-bte`) currently crash at import time with:

```
File ".../nvalchemiops/math/spherical_harmonics.py", line 614, in <module>
    @wp.func
AttributeError: module 'warp' has no attribute 'vec'. Did you mean: 'vec2'?
```

`upet 0.2.2` hard-pins `nvalchemi-toolkit-ops==0.2.0`, whose `spherical_harmonics` module annotates kernels with the generic `wp.vec(length=N, dtype=...)` constructor. NVIDIA Warp removed that alias in `warp-lang 1.10` (only the typed `wp.vec2` / `wp.vec3` / ... remain), so warp's `@wp.func` decorator's annotation evaluation explodes when the module is imported via `metatomic.torch.ase_calculator`. `nvalchemi-toolkit-ops 0.3.0+` switched these kernels to `vec5d` / `vec9d`, which work on `warp-lang 1.13`.

`upet 0.2.1` leaves `nvalchemi-toolkit-ops` unpinned, so requesting `nvalchemi-toolkit-ops>=0.3.0` is enough to make pip backtrack to `upet 0.2.1` + `nvalchemi-toolkit-ops 0.3.1` + `warp-lang 1.13` (resolved on rg.cosmolab as `upet-0.2.1`, `nvalchemi-toolkit-ops-0.3.1`, `warp-lang-1.13.0`, `scipy-1.16.3`, `torch-2.10.0`).

A follow-up upstream fix would be to release a `upet 0.2.3` that pins `nvalchemi-toolkit-ops>=0.3.0` (or unpins it altogether like 0.2.1 did). This is just the cookbook-side workaround.

PR #254, opened 2026-05-04
Update upet everywhere

pfebrer:
Updates upet everywhere with the new release instead of the temporal fix in #253


HaoZeke (approved):
LGTM, maybe we should have the `0.2.2` release yanked though

HaoZeke (approved):
LGTM, maybe we should have the `0.2.2` release yanked though

PR #258, opened 2026-05-12
Create a utility package to be imported in the recipes

ceriottm:
Streamline recurring operations, ATM:
1. download with retry
2. running a shell command

For the moment (2) hasn't been used, left for a second PR. Recipes won't work because they expect the package to be on pipy so we need to publish a first time manually before converting this to a full PR.

### DIFF
--- a/.github/workflows/publish-utils.yml
+++ b/.github/workflows/publish-utils.yml
@@ -1,0 +1,84 @@
+# Build (and on tag, publish) the atomistic-cookbook-utils package.
+#
+# - On pushes to any `utils-v*` tag: build sdist + wheel and publish to PyPI
+#   using Trusted Publishing (OIDC). Configure the publisher at
+#   https://pypi.org/manage/account/publishing/ pointing at this workflow
+#   and the `pypi` environment.
+# - On pull requests that touch the package or this workflow: build only
+#   (no upload) as a smoke test.
+name: Publish atomistic-cookbook-utils
+
+on:
+  push:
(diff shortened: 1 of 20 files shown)


pfebrer:
You didn't push any code for the utils, no?

I'm worried that the package might be a bit overkill and complicate things unnecessarily:
- The download with retry function is probably worth it, but the `run_command` function is arguably not very useful, since one can do `subprocess.run("mtt train options.yaml", shell=True)`.
- Every time we update a util we need to do a release?

pfebrer:
Maybe we can discuss this one in the next ml-devel meeting?

ceriottm:
forgot to add. files are there, you can have a look. it's pretty uncontroversial IMO, and there are other things we could simplify once this is available.

ceriottm:
> You didn't push any code for the utils, no?
> 
> I'm worried that the package might be a bit overkill and complicate things unnecessarily:
> 
> * The download with retry function is probably worth it, but the `run_command` function is arguably not very useful, since one can do `subprocess.run("mtt train options.yaml", shell=True)`.

I think the run_sh() could also do a few extra things like capturing the output by default and printing it back so it gets added to the following cell. One can do without, of course but it's helping

> * Every time we update a util we need to do a release?

Yes, but I don't think we would update utils very often

ceriottm:
Added doc page

pfebrer:
Ok, thanks, the package is on pypi now :+1:

PR #260, opened 2026-05-17
Use ac-utils in recipes for downloads ad shell commands

ceriottm:
As the title says, make use of the newly released atomistic-cookbook-utils

### DIFF
--- a/examples/batch-cp2k/environment.yml
+++ b/examples/batch-cp2k/environment.yml
@@ -6,4 +6,4 @@ dependencies:
   - pip:
     - ase
     - matplotlib
-    - requests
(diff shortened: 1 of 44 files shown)


pfebrer (approved):
There are some `os.makedirs` calls that have been removed, does the new util do this automatically? Anyway I imagine if there are no failures it is not a problem

pfebrer:
Ok I see that indeed the util creates the directory, all good 👍

ceriottm:
> Ok I see that indeed the util creates the directory, all good 👍

See? removing 180 lines of boilerplate ^_^

Issue #228, opened 2026-02-22
nox -e _recipe_name can crash with JSONDecodeError if conda prints warnings


venkatkapil24:
It runs conda _____  --json --dry-run and expects clean JSON, but if Conda prints any warning, nox mixes warning text with JSON and json.loads() fails. The error message is generic and doesn't point to an underlying conda "issue".  

The fix is easy. Add to the following line

https://github.com/venkatkapil24/atomistic-cookbook/blob/7dc41a2a0cbde10b00609d0b14cc5b29b1974739/noxfile.py#L363

`stderr=subprocess.PIPE,
`
this ensures the output is a json.

venkatkapil24:
@ceriottm do you want me to fix this in #217 ?

pfebrer:
I'm not Michele but I would say please don't mix it with the other PR 😅

Can I ask which warning do you get? Maybe it is better to deal with the warning. But your PR seems to build fine no?

ceriottm:
I also got these kinds of errors when running locally. Updating conda tends to fix them, but indeed maybe it's indeed more user friendly not to fail without giving a meaningful error message.  Should definitely be a separate PR though, and I'd like to intercept the warning/error, make sure it is just a warning and print it out in such a way it does not mess with parsing the json, but that the user sees the warning issued by conda.

Luthaf:
Fixed by #268

PR #268, opened 2026-06-02
Handle Conda warnings during JSON dependency resolution

enikidis:
Addresses #228.

This PR updates `update_dependencies()` so that warning output from the `conda env create --json --dry-run` dependency-resolution step does not interfere with parsing the JSON output.

Conda stderr is now captured separately from stdout. If Conda emits warning text while resolving dependencies, it is reported through the Nox session warning mechanism, while `json.loads()` still receives clean stdout JSON.

Validation:

- Reproduced the original failure locally with a fake Conda executable that emits a warning on stderr and valid JSON on stdout.
- Confirmed that after the fix, the warning is shown separately and `update_dependencies()` succeeds.
- Ran `python -m py_compile noxfile.py`.
- Ran `nox -e water-model -- --no-website`; the session completed successfully.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -6,6 +6,7 @@ import re
 import shutil
 import subprocess
 import sys
+import tempfile
 import zipfile
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


ceriottm:
Thanks. Looks good, can you run the linter so it clears that problem? All the warp_ errors are courtesy of nvalchemy, we'll fix those separately and rebase.

enikidis:
Thanks. I applied the formatting fix and pushed an updated commit. Please let me know if there is anything else I should adjust.

ceriottm (approved):
Works and solves the problem

PR #277, opened 2026-06-29
Trying to fix the water model example

pfebrer:
I think the error is due to using an old version of metatensor-torch


pfebrer:
I think this one might need updating torch-pme @PicoCentauri @sirmarcel

PicoCentauri (approved):

PR #278, opened 2026-06-29
Limit metatrain version in eon-pet-neb

pfebrer:
I think `eon` does not yet support the new metatomic/metatensor, am I right @HaoZeke ?


HaoZeke:
yup, working on it, will need to do a release

pfebrer:
Solved in #280

PR #285, opened 2026-07-07
Try to fix llpr example

pfebrer:
Since the introduction of cutoff stats in PET, one can't ask for all supported outputs to be written to an xyz file until https://github.com/metatensor/metatrain/issues/1185 is solved


pfebrer:
Ok I think it's not possible to fix it like this, the ASE writing will need to be fixed

PR #287, opened 2026-07-10
Fix non-degenerate TA branches in the Si dispersion

gbarbalinardo:
## Summary

- Fix non-degenerate transverse acoustic branches near Gamma in the Si dispersion
- `is_unfolding=True` on Phonons: uses Wigner-Seitz folding so degeneracies hold at q-points incommensurate with the 3x3x3 supercell (also fixes the BTE, whose 5x5x5 mesh is incommensurate)
- `is_acoustic_sum=True` on ForceConstants: zeroes the Gamma acoustic modes
- Force-constant symmetrization is now on by default in kaldo, so the PET-MAD symmetry breaking near Gamma is projected out
- Requires kaldo>=2.2.0

### DIFF
--- a/examples/thermal-conductivity-bte/environment.yml
+++ b/examples/thermal-conductivity-bte/environment.yml
@@ -7,7 +7,7 @@ dependencies:
     - ase>=3.23
     - upet>=0.2.4
     - chemiscope>1.0
-    - kaldo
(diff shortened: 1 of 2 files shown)


pfebrer:
Thanks for the fix!

PR #290, opened 2026-07-16
Pimd mad example

ceriottm:
This is just adding a small example of using a MLIP for a PI simulation. Would need to merge it VERY fast as I want to use it for a summer school in three days ^_^


examples/path-integrals/path-integrals.py
```diff
@@ -391,3 +394,110 @@
...
+    xml_content = file.read()
+print(
+    xml_content[
+        xml_content.find("<ffdirect") : xml_content.find("</ffdirect>")
+        + len("</ffdirect>")
+    ]
+)
+
+ipi_process = subprocess.Popen(["i-pi", "data/input_pet_piglet.xml"])
+ipi_process.wait()
```
pfebrer: No atomistic cookbook utils? 👀

ceriottm: point well taken. should add a mechanism to also run in the background, will do in a separate PR

PR #295, opened 2026-07-20
A recipe on vibrations from PI-approximations with i-PI

mahrossi:
This is the adaptation of the latest tutorial.


mahrossi:
Hey @pfebrer - trying to contribute and the CI is failing on generating the doc for an unrelated repo. Do you know about this? Seems like it cannot run some gromacs simulations -- expected? Or do you know what to do?

Luthaf:
This is unrelated, you can ignore it =). 

@HaoZeke could it be related to today's update of the gromacs-metatomic build?

pfebrer:
Yes when you touch some file outside of the example, you trigger the CI of all examples in the cookbook :) But as Guillaume says, in this case it can be ignored, it will be fixed elsewhere.

Let me know when it is ready for review :+1:

mahrossi:
Great, it is ready for review

mahrossi:
- Items to do (and to discover how to do) after receiving feedback from @ceriottm :

    - Files are only a 150kb each but still too large - need to move to zenodo and do automatic fetch, etc. I think unfortunately even the figures in the page that explain the workflow are too large (around 60kb each), so I am not sure how to make them smaller without making them crappy, but will try. Goal is to have the whole repository at a few 10s of kb.
    
    - Questions are not in the right style for this repository. Questions and interpretation should be removed or discussed differently, as this should not resemble an exercise. Move questions and discussion to some other point or not here at all. Just present setup and how to run (if I understood right).
    
    
-> can't say I have time soon but I try at some point!

mahrossi:
ok I think this is more cookbookish and data is fetched from zenodo, making it really small. 

I would still like to do one improvement, namely run (launch) one short simulation with all methods from the notebook. However, I will do that as a second improvement stage. 

Check if it can be merged at the current state.

pfebrer:
Thanks! I reverted the changes outside the example since they are not needed anymore, and this will allow the website to be built so that I can check it :+1:

pfebrer (commented):
Looks very good to me!

However, if the i-PI simulations are short to run it would be nice indeed do run them in the cookbook example. This ensures that what is proposed in the cookbook works, and it warns us when some package update makes it fail, so that we can fix it.

I can try to add this tomorrow if you want from the instructions that you give in the tutorial, it should not be hard. Unless of course these simulations take a long time, let me know if that is the case.

mahrossi:
Thanks @pfebrer -- I believe one "run" should take no more than 1 minute -- the most expensive ones are the "CMD" ones that use a small time step. The reference that is downloaded from zenodo does 10 runs for each method (6 or 7 methods, so 70 trajectories), so that would really be probably a bit more than 10 mins.

Let me know if you can easily do it - if not, I take it on me in the coming weeks.

pfebrer:
Hmm we have some other examples that take around 10min so it could be fine if we can keep it to 10 or a bit less. Is there the possibility to reduce run time while still getting reasonable results? It would be fine if we do the plots from the cookbook-generated data and then say: "We also provide more converged data to plot, one can download it from Zenodo".

In general the cookbook recipes should run the full workflow, even if they can't get to the most converged results. The goal is to ensure that the full workflow is reproduceable.

mahrossi:
Hm, IDK - depending on where this is running it could be quite over 10 mins. Sure it is possible to plot one or two trajectories (even though you would have to code the opening of several directories and different starting points, etc. -- which is a bit much maybe for a cookbook?) - not super sure. Also you would create quite extensive outputs for the 70 or 80 runs -- is it ok to store lots of data in whatever place this is running? I give only the "already worked out data" which is mush smaller in size.

mahrossi:
But you can of course run and plot 1 trajectory for each method within the cookbook - I predict that might take less than 2 minutes (about 1 for each longer one and a factor 10 less for the other ones). however: I would keep the analysis on the zenodo data.

pfebrer:
Yes, I think this is the best thing to do, writing the code for N trajectories with N=2 and then say "if you want converged results do it with N=7, we are going to fetch that data from Zenodo to plot"

> (even though you would have to code the opening of several directories and different starting points, etc. -- which is a bit much maybe for a cookbook?)

I think that is exactly the spirit of the cookbook :) To have all the code needed to exactly reproduce some workflow. Because in the current state there are some things that you should do to get the data that are just mentioned but I wouldn't know how to do 😅 

I have never run an i-PI simulation outside of tutorials but I can give it a try if you don't have time 👍

mahrossi:
Hey @pfebrer and maybe also @ceriottm -- this is all good, but I am really at my capacity to work on this for the coming weeks. So if it is really unacceptable to merge this version and update later, this will stay here for some time -- sorry for that. I think this version is helpful and mergeable though, as I wrote in the beginning, so let me know.

pfebrer:
Up to Michele :) I think it is indeed a very nice tutorial/pedagogical exercise, I'm just trying to interpret the words of the creator on what the cookbook should contain haha So if the creator himself thinks it is good to go, all good!

mahrossi:
Hi everyone, in about a month from now, it would be good to use this cookbook for a new tutorial or at least advertise it. I was beyond very busy in the past month, so I had no time to fix the rest of what is needed. It would be nice to have it up. I can try to ask for help from a student of mine to make this work as you require here, but otherwise I think I am maxed-out. Can someone on your side help?

pfebrer:
I can help guiding a student that works on it, but I don't have time to do the work myself either :sweat_smile:

PR #296, opened 2026-07-28
fix(ml-mm): support gromacs-metatomic mta3 MKL stack

HaoZeke:
## Summary

Documentation CI [run 1981](https://github.com/lab-cosmo/atomistic-cookbook/actions/runs/30252861176) failed on `generate-example (ml-mm)` after the channel moved from `gromacs-metatomic` **mta2** to **mta3**:

```
gmx_mpi: error while loading shared libraries: libmkl_intel_lp64.so.3: cannot open shared object file
```

`mta3` links `libgromacs` against MKL but does not declare `mkl` as a runtime dependency. The recipe still forced `libtorch=cpu_generic`, which pulls `nomkl` and leaves those libraries missing.

This PR:

- installs `mkl` and uses the default MKL-backed `libtorch` / `pytorch-cpu` (shared BLAS with `libgromacs`)
- raises the pip stack to match the published mta3 C++ ABI (`libmetatomic-torch >=0.1.16`, `libtorch >=2.12`, `metatrain>=2026.3`)

## Verification

Reproduced and fixed on a remote host with a clean env from `examples/ml-mm/environment.yml`:

- `gmx_mpi -version` loads
- full `python ml-mm.py` finishes (`grompp` + 2000-step `mdrun` with Metatomic potential + RMSD/Rama/chemiscope analysis)

## Test plan

- [x] Full `examples/ml-mm` recipe on remote host with the new pins
- [ ] CI `generate-example (ml-mm)` green on this PR

### DIFF
--- a/examples/ml-mm/environment.yml
+++ b/examples/ml-mm/environment.yml
@@ -3,12 +3,11 @@ channels:
   - conda-forge
 dependencies:
   - chemiscope
-  # Single-precision OpenMPI only. The mpi_* glob also matches mpi_*_dblprec_*,
-  # which installs gmx_mpi_d and leaves shutil.which("gmx_mpi") empty (main CI).
-  # Build string mpi_openmpi_h* excludes dblprec and cuda variants.
+  # Single-precision OpenMPI build only. The broader mpi_* match also selects
+  # double-precision packages that ship gmx_mpi_d instead of gmx_mpi.
   - metatensor::gromacs-metatomic=*=mpi_openmpi_h*
-  # Feedstock #4 links GROMACS to flexible libblas/liblapack (not MKL sonames).
-  # Default libtorch/pytorch-cpu is fine; do not force cpu_generic/nomkl.
+  # gromacs-metatomic links against flexible libblas/liblapack; default
+  # libtorch/pytorch-cpu is compatible without a forced BLAS vendor pin.
   - libtorch
   - pytorch-cpu
   - matplotlib
@@ -17,8 +16,7 @@ dependencies:
   - pip
   - uv
   - pip:
-    # Align Python metatomic with published mta3 (libmetatomic-torch >=0.1.15,
-    # libtorch >=2.12).
+    # Match published mta3 (libmetatomic-torch >=0.1.15, libtorch >=2.12).
     - torch>=2.12,<2.13
     - ase>=3.23
     - cmcrameri
--- a/examples/ml-mm/environment.yml
+++ b/examples/ml-mm/environment.yml
@@ -4,18 +4,14 @@ channels:
 dependencies:
   - chemiscope
   - metatensor::gromacs-metatomic=*=mpi_*
-  # gromacs-metatomic links libgromacs against libopenblas.so.0 but the recipe
-  # does not list it as a runtime dependency; pin it here so gmx_mpi can load.
-  - libopenblas
-  # Use the cpu_generic libtorch build (linked against openblas) instead of the
-  # default cpu_mkl one. With cpu_mkl, libtorch's mkl_gemm_batched calls
-  # cblas_sgemm_batch, which the dynamic loader resolves to libopenblas (pulled
-  # in by libgromacs) rather than libmkl. The OpenBLAS implementation crashes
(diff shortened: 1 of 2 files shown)


pfebrer:
Why doesn't the feedstock declare mkl as a runtime dependency if it needs it?

HaoZeke:
> Why doesn't the feedstock declare mkl as a runtime dependency if it needs it?

I can't recall now but I think it was some attempt to make it BLAS backend agnostic.. which didn't work so well, I can target the feedstock a bit

HaoZeke:
Follow-up on the feedstock question:

GROMACS free-found MKL from host `libtorch * cpu*` (cpu_mkl) and recorded `DT_NEEDED` on `libmkl_*.so.3` in `libgromacs` for its own Fortran BLAS (`daxpy_`, `dgemm_`, …). The recipe never declared `mkl` (or any BLAS) as a direct run dep — the “backend agnostic” attempt was “libtorch will pull the right BLAS”, which breaks when a client re-pins `libtorch=cpu_generic` / `nomkl`.

Feedstock fix (flexible `libblas`/`liblapack` instead of hard-coding MKL): https://github.com/metatensor/gromacs-metatomic-feedstock/pull/4

This cookbook PR remains the client-side workaround until that rebuild is on the channel.

pfebrer:
We don't need this anymore no? main should work already (?)

pfebrer:
Or at least not the mkl thing

Luthaf:
Yes, there is a new conda build that fix the mkl issue, and another one coming in https://github.com/metatensor/gromacs-metatomic-feedstock/pull/5. Once this is merged, we can rebuild here and everything should be fine

Luthaf:
Now that we have new builds on conda, is this still required? and/or can it be simplified?

examples/ml-mm/environment.yml
```diff
@@ -3,33 +3,24 @@ channels:
...
-  # because it is being called with MKL's argument layout (SIGSEGV inside
-  # inner_small_matrix_thread). Forcing cpu_generic makes both libtorch and
-  # libgromacs share the same openblas BLAS, eliminating the symbol clash.
-  - libtorch =*=cpu_generic*
-  - pytorch-cpu =*=cpu_generic*
+  # Single-precision OpenMPI build only. The broader mpi_* match also selects
+  # double-precision packages that ship gmx_mpi_d instead of gmx_mpi.
+  - metatensor::gromacs-metatomic=*=mpi_openmpi_h*
+  # gromacs-metatomic links against flexible libblas/liblapack; default
+  # libtorch/pytorch-cpu is compatible without a forced BLAS vendor pin.
```
pfebrer, line 10: No need to comment this I would say

pfebrer: (same for the comment before this one)

examples/ml-mm/environment.yml
```diff
@@ -3,33 +3,24 @@ channels:
...
-  # default cpu_mkl one. With cpu_mkl, libtorch's mkl_gemm_batched calls
-  # cblas_sgemm_batch, which the dynamic loader resolves to libopenblas (pulled
-  # in by libgromacs) rather than libmkl. The OpenBLAS implementation crashes
-  # because it is being called with MKL's argument layout (SIGSEGV inside
-  # inner_small_matrix_thread). Forcing cpu_generic makes both libtorch and
-  # libgromacs share the same openblas BLAS, eliminating the symbol clash.
-  - libtorch =*=cpu_generic*
-  - pytorch-cpu =*=cpu_generic*
+  # Single-precision OpenMPI build only. The broader mpi_* match also selects
+  # double-precision packages that ship gmx_mpi_d instead of gmx_mpi.
```
pfebrer, line 7: And that would be a problem?

pfebrer:
Yes I would only add the minimal comments and as non-technical as possible, since the environnent.yml is something that the users are meant to see and lots if technical comments might confuse them

examples/ml-mm/environment.yml
pfebrer: (lines 9 and 10)

PR #302, opened 2026-09-01
Fix the pimd-boson recipe taking too long

DavideTisi:
this fix the pimd-boson 
i created the PR to trigger the CI


DavideTisi:
Hi @BarakHirshberg, I have just cut the durations of the simulations in half; it passes the CI. You can check if the results are still reasonable by downloading the zip file.

BarakHirshberg:
Hi @DavideTisi , I checked and it still looks decent with the reduced number of steps. Thank you for taking care of it.

Luthaf:
thanks @DavideTisi!

